PR #16, opened 2020-07-13
Add mapK to Span

catostrophe:
I had one for `natchez.Span` but during migration to `trace4cats` I found out it couldn't be implemented as an extension because of protected methods.
Thus, I am adding it as a member of the Span itself.


janstenpickle:
Looks good, thanks for the contribution!

janstenpickle:
@catostrophe I'll do a 0.3.0 release later this week. In the meantime you should be able to use version `0.2.0+6-29c0a493` which has your changes included.

Thanks again!

PR #18, opened 2020-07-16
Add noop EntryPoint

### LINKED ISSUE #17: The need for no-op entities
In order to make tracing optional (switchable) at the application level, e.g. via config, we need to be able to create no-op instances of some entities.

We have one for `Trace`. But the one for `EntryPoint` is missing. As to implement it, we need a `NoopSpan` which in turn cannot be created without the empty `SpanContext`.

I will add those listed above if we define what the empty SpanContext is.

catostrophe:
Closes #17


janstenpickle:
Thanks!

janstenpickle:
@catostrophe This should be available in the `0.2.0+10-22fefa0f` build

PR #30, opened 2020-07-30
[http4s] Make span operation name configurable

### LINKED ISSUE #23: [http4s] Using req.uri.path as an operation name spoils the operation namespace.
"The name should be low cardinality, so it should not include identifiers."
Using the raw path is not good. Also, it's already included in the span as the `http.url` tag.

catostrophe:
Closes #23


catostrophe:
@janstenpickle the motivation for making `spanNamer` function existential is because:
1. user should be able to provide it as a simple lambda.
2. it would be a mistake to parameterize it with the same F[_] as in HttpRoutes/Client. Trying to implement `io.janstenpickle.trace4cats.http4s.client.ClientSyntax.TracedClient#inject` demonstrates it clearly.
3. polymorphic functions aren't in Scala 2, and even in Scala 3 type inference is not good.

The build has just broken on Scala 2.12. Trying to fix it.

PR #127, opened 2020-11-17
TraceHeaders wrapper and converters for integrations

### LINKED ISSUE #124: Introduce a newtype for headers map used in EntryPoint and ToHeaders
I would like to have a newtype wrapping `Map[String, String]` for headers being passed to `EntryPoint`'s and `ToHeaders`'s methods as [it's done in Natchez](https://github.com/tpolecat/natchez/blob/master/modules/core/shared/src/main/scala/Kernel.scala).

Something simple like the following would be fine:
```scala
final case class TraceHeaders(toMap: Map[String, String]) extends AnyVal
```

I would do it myself, if you're ok with the naming. Which module should it be put into. I suppose `core`?

The motivation is that in my codebase I write custom span injectors and I require optics for abstract type parameters, which parameterize endpoint inputs or evaluation context, and for optics, especially implicit ones, it is highly desirable to have newtypes. That is `Getter[Ctx, TraceHeaders]` is better than `Getter[Ctx, Map[String, String]]`.

I use it in my code, but I think it also may be useful in the library.

catostrophe:
Closes #124 

Along with `TraceHeaders` type itself, I also added converters to/from integrations to remove in-place conversions.

@janstenpickle need your review on naming

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -526,7 +526,7 @@ lazy val `sttp-client` = (project in file("modules/sttp-client"))
       Dependencies.sttpHttp4s
     )).map(_ % Test)
   )
-  .dependsOn(model, kernel, core, inject, `exporter-common` % "test->compile")
+  .dependsOn(model, kernel, core, inject, `exporter-common` % "test->compile", `http4s-common` % "test->compile")
 
 lazy val `http4s-client` = (project in file("modules/http4s-client"))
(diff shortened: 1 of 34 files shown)


janstenpickle (approved):
This is great, thanks very much @catostrophe!

PR #131, opened 2020-11-18
Fix SttpHeaders.converter

catostrophe:
This is to fix a bug in sttp headers conversion, appeared in #127


janstenpickle (approved):
Nice catch, thanks 👍

PR #132, opened 2020-11-18
Add aliases for global scalafmt formatting

catostrophe:
I hate when my PRs fail due to `scalafmtCheck`. 

Add `sbt` aliases that globally check and format the whole codebase with scalafmt:
* `fmt` formats all sources in all modules
* `fmtCheck` checks the same

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -733,3 +733,6 @@ lazy val `collector-lite` = (project in file("modules/collector-lite"))
     `graal-kafka`
   )
   .enablePlugins(GraalVMNativeImagePlugin)
+
+addCommandAlias("fmt", "all root/scalafmtSbt root/scalafmtAll")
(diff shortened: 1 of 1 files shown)


janstenpickle (approved):

janstenpickle:
@catostrophe I don't know which IDE you use, but in Intellij there is an option to format with scalafmt on save

catostrophe:
@janstenpickle I use IntelliJ and that setting is on, but for some reason, my builds tend to fail on scalafmtCheck.

janstenpickle:
@catostrophe oh that's annoying, here's my settings just in case there's something different with your setup: 
Screenshot 2020-11-18 at 15 39 02

catostrophe:
@janstenpickle yeah, the same as mine.

PR #155, opened 2020-12-02
Generalise entrypoint

janstenpickle:
introduce new typeclasses for creating a generalised
entrypoint. Allows some environment `R` to be created
from some other environment `R0`

### DIFF
--- a/modules/base/src/main/scala/io/janstenpickle/trace4cats/base/context/Init.scala
+++ b/modules/base/src/main/scala/io/janstenpickle/trace4cats/base/context/Init.scala
@@ -1,0 +1,34 @@
+package io.janstenpickle.trace4cats.base.context
+import cats.Monad
+
+trait Init[Low[_], F[_], R, R0] extends Provide[Low, F, R] {
+  def P: Provide[Low, F, R]
+
+  def init[A](fa: F[A])(r0: R0): Low[A]
+
+  override def provide[A](fa: F[A])(r: R): Low[A] = P.provide(fa)(r)
+
+  override def ask[R1 >: R]: F[R1] = P.ask
+
+  override def lift[A](la: Low[A]): F[A] = P.lift(la)
+
+  override def local[A](fa: F[A])(f: R => R): F[A] = P.local(fa)(f)
+}
+
+object Init {
+  def apply[Low[_], F[_], R, R1](implicit init: Init[Low, F, R, R1]): Init[Low, F, R, R1] = implicitly
+
+  implicit def fromInjectProvide[Low[_]: Monad, F[_]: Monad, R, R0](implicit
+    I: Inject[Low, R, R0],
+    pv: Provide[Low, F, R]
+  ): Init[Low, F, R, R0] =
+    new Init[Low, F, R, R0] {
+      override def P: Provide[Low, F, R] = pv
+
+      override def init[A](fa: F[A])(r0: R0): Low[A] = I(r0)(P.provide(fa))
+
+      override def Low: Monad[Low] = implicitly
+
+      override def F: Monad[F] = implicitly
+    }
+}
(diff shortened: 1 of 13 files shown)


janstenpickle:
@catostrophe this is attempting to generalise the `makeContext` parameter you added to the http4s syntax

modules/base/src/main/scala/io/janstenpickle/trace4cats/base/context/Inject.scala
```diff
@@ -0,0 +1,12 @@
+package io.janstenpickle.trace4cats.base.context
+
+trait Inject[F[_], R, R0] { self =>
```
catostrophe, line 3: What is the common sense behind the `Inject` trait? I think you mixed 2 different concepts here.

As I see it, the aim of `Inject` is to encapsulate only one function: `R0 => G[R]` where `G[_]` is something usable: typically a resource over F, or (rarely) is F itself.

So given instances 
```scala
val inject: Inject[F, R, R0] = ???
val r0: R0 = ???
val f: R => F[A] = ???
```
a call to inject:
```scala
inject.apply(r0)(f)
```
may be replaced with
```scala
makeR(r0).use(f)
```
where
```scala
def makeR: R0 => Usable[R] = ???  //this is what an `Inject` instance actually encapsulates
```
where `Usable[R]` is `Resource[F, R]` or (rarely) `F[R]` with use=flatMap.

We can even introduce a new trait:
```scala
trait Usable[F[_], A] {
  def use[B](f: A => F[B]): F[B]
}

object Usable {
  def fromEffect[F[_]: Monad, A](res: F[A]): Usable[F, A] =
    new Usable[F, A] {
      def use[B](f: A => F[B]): F[B] = res.flatMap(f)
    }

  def fromResource[F[_]: BracketThrow, A](res: Resource[F, A]): Usable[F, A] =
    new Usable[F, A] {
      def use[B](f: A => F[B]): F[B] = res.use(f)
    }
}
```

But I think it's overkill and we can just go ahead with the concrete `Resource[F, A]`, because we use it for spans only, and a span is always a resource.

I think the second part of the curried `apply`(`(f: R => F[A]): F[A]`)  is redundant here. And only the first one is meaningful. So in fact, `Inject` is just a resource provider. And the resource itself has the ability to be used with a function `f`.

So, does a simple provider function deserve a separate typeclass?
What about 
```scala
type SpanProvider[F[_], R] = R => Resource[F, Span[F]]
```

modules/base/src/main/scala/io/janstenpickle/trace4cats/base/context/Init.scala
```diff
@@ -0,0 +1,36 @@
+package io.janstenpickle.trace4cats.base.context
+import cats.{~>, Monad}
+
+trait Init[Low[_], F[_], R, R0] extends Provide[Low, F, R] {
```
catostrophe, line 4: As well, I don't see any need to combine a resource provider with the `Provide` typeclass.

modules/base/src/main/scala/io/janstenpickle/trace4cats/base/context/Inject.scala
catostrophe: With all said above, I think both `Inject` and `Init` are not simple typeclasses, but some overcomplicated constructions that may be divided into more simple things. Moreover, they don't belong to the `context` package or maybe even the `base` module.

The original idea behind the context package was to create typeclass hierarchy for contextual effects F[_], such that forall `A`: `F[A]` is isomorphic to the `R => Low[A]` arrow. Among them are `cats.data.Kleisli`, `zio.ZIO`, `tofu.env.Env`, etc.

janstenpickle: Thanks @catostrophe this all makes sense. I had my reservations about making a typeclass too, especially as, you pointed out, it behaves more as a constructor. I will try and do something more in `inject` which can be used with the entrypoint.

PR #157, opened 2020-12-03
Generalise entrypoint

janstenpickle:
- Allows a context to be created from some runtime R
- `EntryPoint` may be transformed into `ContextConstructor`
- Paves the way for different context support in FS2

PR #164, opened 2020-12-08
Add syntax for Trace[F]

catostrophe:
Will conflict with #157. Need to resolve.


catostrophe:
```
BindException was thrown during property evaluation.
[info]     Message: Address already in use
```

catostrophe:
@janstenpickle restart the job please

PR #177, opened 2020-12-18
Remove Child links, leave a single Link class for all link types

### LINKED ISSUE #151: Wrong SpanRefType mapping in JaegerSpanExporter
I think there is an error in the mapping between the internal `trace4cats` span model and the Jaeger model.

https://github.com/janstenpickle/trace4cats/blob/82baa78bd7ee1c5e7742d99d4f53477a43bd76f7/modules/jaeger-thrift-exporter/src/main/scala/io/janstenpickle/trace4cats/jaeger/JaegerSpanExporter.scala#L76-L77

Here's the link to an interesting discussion regarding their meaning:
https://github.com/open-telemetry/opentelemetry-specification/issues/65

They both mean "parent" or "ancestor", not a child, but in a bit different way.

The below quote is especially interesting:
> Another odd thing about child-of and follows-from is that it's the child span that defines this reference type, even though it talks about parent's dependency on child outcome. If you're a remote server, how do you even know if parent/caller does or does not depend on your outcome? I tend to think of this as the nature of the protocol: producer of a message to Kafka does not respect any response, so the receiver should use follows-from. Sender of HTTP request does expect a response, so the server always uses child-of, even of the sender doesn't care about the outcome - in that case it can internally create a follows-from span first, and then a normal pair of RPC call spans. So it's possible to rationalize this way, but it's still kind of dirty.  

I suggest changing it to:
```scala
case Link.Parent(_, _) => new SpanRef(SpanRefType.CHILD_OF, traceIdLow, traceIdHigh, spanId) 
```
That is when we add a link `Parent(parentTraceId, parentSpanId)` we mean that the current span is a **child of** (parentTraceId, parentSpanId).

catostrophe:
Closes #151

### DIFF
--- a/modules/jaeger-integration-test/src/main/scala/io/janstenpickle/trace4cats/test/jaeger/BaseJaegerSpec.scala
+++ b/modules/jaeger-integration-test/src/main/scala/io/janstenpickle/trace4cats/test/jaeger/BaseJaegerSpec.scala
@@ -79,12 +79,7 @@ trait BaseJaegerSpec extends AnyFlatSpec with ScalaCheckDrivenPropertyChecks wit
                         JaegerReference("CHILD_OF", traceId.show, parent.spanId.show)
                       } ++ span.links
                         .fold(List.empty[JaegerReference])(_.map { link =>
-                          val linkType = link match {
-                            case Link.Child(_, _) => "CHILD_OF"
-                            case Link.Parent(_, _) => "FOLLOWS_FROM"
-                          }
-
-                          JaegerReference(linkType, link.traceId.show, link.spanId.show)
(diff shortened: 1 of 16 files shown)


catostrophe:
Ouch 😰
This happened again:
```
BindException was thrown during property evaluation.
[info]     Message: Address already in use
```

The port is used only once in the codebase. But the tests sometimes fail.

@janstenpickle could you restart the job

janstenpickle:
Oh man, that's getting really annoying, I think I'll change the tests so we just send the request straight to the routes Kleisli.

janstenpickle (approved):

PR #187, opened 2020-12-21
Add natural tranformations to base

janstenpickle:


catostrophe:
@janstenpickle Why did you add mapK/imapK only to 2 of them? Do you have any particular application? Could they be useful in Ask/Local?

janstenpickle:
> @janstenpickle Why did you add mapK/imapK only to 2 of them? Do you have any particular application? Could they be useful in Ask/Local?

It's because the inhertance structure causes issues when adding these methods on the less powerful abstractions.

catostrophe:
> > @janstenpickle Why did you add mapK/imapK only to 2 of them? Do you have any particular application? Could they be useful in Ask/Local?
> 
> It's because the inhertance structure causes issues when adding these methods on the less powerful abstractions.

Hmmm... I think I can do it. Let me try tonight.

catostrophe:
I can't push to PR branches, so I created a separate PR #191

PR #189, opened 2020-12-22
Add integration with sttp-client3 and tapir

### LINKED ISSUE #128: Sttp integration module needs reorganization
I suggest some changes:

- [x] rename `backend` package to `client` as it's named in `sttp`
- [x] consider support for `sttp.client3`
- [x] add integration with `tapir`: `ServerEndpoint` may be span injected (in many cases it is preferable over routes tracing)
- [x] add `common` package for common code used for sttp client and tapir integrations

catostrophe:
Sttp `client3` and `tapir` have a common model, incompatible with `client` v2. Those using `tapir` should also use `client3`. 

Resolves #128


janstenpickle (approved):
Nice, thanks for keeping sttp client 2!

PR #261, opened 2021-02-09
Update .mergify.yml

catostrophe:
Automatically merge dependabot's PRs.

Add `dependencies` label for Steward's PRs - it may be useful to mark all dependency updates with this label.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,21 +1,66 @@
 pull_request_rules:
-  - name: automatically merge scala-steward's PRs
+  - name: delete head branch after merge
+    conditions: []
+    actions:
+      delete_head_branch: {}
+
+  - name: request reviews and label scala-steward's PRs
+    conditions:
+      - author=scala-steward
+    actions:
+      request_reviews:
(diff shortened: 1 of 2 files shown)


catostrophe:
Probably we may also change the config to auto-merge those Steward's PRs that update **major/minor** version, passed all tests, and got **1 approval**.

.mergify.yml
```diff
@@ -1,10 +1,66 @@
...
-        method: merge
\ No newline at end of file
+        method: merge
+
+  - name: automatically merge other scala-steward's PRs reviewed by 1 maintainer
+    conditions:
+      - author=scala-steward
+      - status-success~=test \(2.12
+      - status-success~=test \(2.13
+      - "#approved-reviews-by>=1"
```
catostrophe, line 53: hm... or maybe "#approved-reviews-by>=2"

janstenpickle: I think this is fine as 1 for now, if we add more maintainers in the future let's increase it.

PR #273, opened 2021-02-11
Wire span error handler through to inject module

janstenpickle:
The error handler on span in the core module wasn't used in the higher level abstractions.

@catostrophe I was thinking of adding something to the error handler here to set an attribute on error as well, what do you think?


modules/sttp-tapir/src/main/scala/io/janstenpickle/trace4cats/sttp/tapir/TapirResourceKleislis.scala
```diff
@@ -32,9 +33,10 @@ object TapirResourceKleislis {
     inHeadersGetter: Getter[I, Headers],
     inSpanNamer: TapirInputSpanNamer[I],
     errorToSpanStatus: TapirStatusMapping[E],
-    dropHeadersWhen: String => Boolean = HeaderNames.isSensitive
+    dropHeadersWhen: String => Boolean = HeaderNames.isSensitive,
+    errorHandler: PartialFunction[Throwable, SpanStatus] = PartialFunction.empty,
```
catostrophe, line 37: Tapir injector has already got some sort of `errorToSpanStatus` mapping. I need to look at it carefully tomorrow.

janstenpickle: Yeah, good point. So does the http4s injector. I don't think we actually need this here, but child spans in a user's application can still benefit from it I think.

modules/http4s-server/src/main/scala/io/janstenpickle/trace4cats/http4s/server/Http4sResourceKleislis.scala
```diff
@@ -1,35 +1,49 @@
...
     requestFilter: Http4sRequestFilter = Http4sRequestFilter.allowAll,
     dropHeadersWhen: CaseInsensitiveString => Boolean = Headers.SensitiveHeaders.contains,
+    errorHandler: ErrorHandler = ErrorHandler.empty,
   )(k: ResourceKleisli[F, SpanParams, Span[F]]): ResourceKleisli[F, Request_, Span[F]] =
     Kleisli { req =>
       val filter = requestFilter.lift(req).getOrElse(true)
       lazy val headers = Http4sHeaders.converter.from(req.headers)
-      val spanResource = if (filter) k.run((spanNamer(req), SpanKind.Server, headers)) else Span.noop[F]
+      val spanResource =
+        if (filter) k.run((spanNamer(req), SpanKind.Server, headers, errorHandler.orElse(messageFailureHandler)))
```
janstenpickle, line 45: @catostrophe what do you think of the fallback precidence here? If someone creates a partial function handling all `Throwable`s it will override the message failure handler.

catostrophe: @janstenpickle I think this is fine since it's overridable and under the control of a user.

catostrophe (approved):

PR #309, opened 2021-02-23
Remove redundant type evidence on collector

janstenpickle:


catostrophe:
We also need to fix #310

catostrophe (approved):

PR #314, opened 2021-02-24
Simplify batching logic of spans

CremboC:


modules/exporter-common/src/main/scala/io/janstenpickle/trace4cats/export/QueuedSpanCompleter.scala
```diff
@@ -27,30 +27,32 @@ object QueuedSpanCompleter {
...
+      _ <- Resource.make {
+        queue.dequeue
+          .groupWithin(batchSize, batchTimeout)
+          .map(spans => Batch(spans))
+          .evalMap { batch =>
+            Stream
+              .retry(
+                exporter.exportBatch(batch).onError { case th =>
+                  Logger[F].warn(th)("Failed to export spans")
+                },
```
janstenpickle, line 43: I think this `onError` would be better outside of the retry now.

modules/exporter-common/src/main/scala/io/janstenpickle/trace4cats/export/QueuedSpanCompleter.scala
```diff
@@ -27,30 +27,32 @@ object QueuedSpanCompleter {
...
+                delay = 5.seconds,
+                nextDelay = _ + 1.second,
```
janstenpickle, line 45: It might be a good idea to make these configurable now and possibly pick some different defaults, what do you think @CremboC?

CremboC: Yeah I was thinking exactly the same. Just wasn't sure what good defaults are, the current ones aren't ideal either I don't think .

janstenpickle: Yeah, maybe intital delay of 500ms, with a backoff of 100ms? To expose this further up it will need wiring through all of the exporters, so if you don't want to do that it's OK by me.

modules/exporter-common/src/main/scala/io/janstenpickle/trace4cats/export/QueuedSpanCompleter.scala
CremboC: I don't think it's ever going to be called outside the retry because of the `maxAttempts = Int.MaxValue`?

modules/exporter-common/src/main/scala/io/janstenpickle/trace4cats/export/QueuedSpanCompleter.scala
CremboC: Addressed in 605a27a

modules/exporter-common/src/main/scala/io/janstenpickle/trace4cats/export/QueuedSpanCompleter.scala
janstenpickle: Yeah, good point. Thinking about it some more, I don't think this should repeat indefinitely, maybe 10 times at most (eventually making it configurable)

PR #324, opened 2021-03-02
Migrate to Cats Effect 3 ecosystem

### LINKED ISSUE #299: Migrate to CE3
This issue is to track dependencies and blockers.

CE3 tracking issue: typelevel/cats-effect#1330

PR with CE3 changes: #374

Dependencies affected by CE3 migration:

- [x] cats-effect - released [v3.1.0](https://github.com/typelevel/cats-effect/releases/tag/v3.1.0) - ok 🟢
- [x] fs2 - released [v3.0.2](https://github.com/typelevel/fs2/releases/tag/v3.0.2) - ok 🟢
- [x] log4cats - released [v2.1.0](https://github.com/typelevel/log4cats/releases/tag/v2.1.0) - ok 🟢
- [x] natchez - released [v0.1.2](https://github.com/tpolecat/natchez/releases/tag/v0.1.2) - ok 🟢 
- [x] decline - released [v2.0.0](https://repo1.maven.org/maven2/com/monovore/decline-effect_2.13/2.0.0/) - ok 🟢 
- [x] zio-interop-cats - released [v3.0.2.0](https://github.com/zio/interop-cats/releases/tag/v3.0.2.0) - ok 🟢 
- [x] sttp-client3 - released [v3.3.2](https://github.com/softwaremill/sttp/releases/tag/v3.3.2) - ok 🟢 
- [x] fs2-kafka - candidate [v2.0.0](https://github.com/fd4s/fs2-kafka/releases/tag/v2.0.0) - ok 🟢 
- [x] redis4cats - candidate [v1.0.0-RC3](https://github.com/profunktor/redis4cats/releases/tag/v1.0.0-RC3) - pending 🟡 
- [x] http4s - milestone [v1.0.0-M21](https://github.com/http4s/http4s/releases/tag/v1.0.0-M21) - pending 🟡 
- [x] http4s-jdk-http-client - milestone [v0.5.0-M4](https://github.com/http4s/http4s-jdk-http-client/releases/tag/v0.5.0-M4) - pending 🟡 
- [ ] tapir-http4s-server - WIP softwaremill/tapir#1154 - blocker 🔴

catostrophe:
Closes #299 

This is going to be a very long-lasting WIP PR...

I am going to go through all the modules that already have got CE3-compatible dependencies and make the code compile.
Then fix the tests. Then wait for the rest dependencies and do the same.


modules/core/src/test/scala/io/janstenpickle/trace4cats/SpanSpec.scala
```diff
@@ -18,10 +18,11 @@ import scala.concurrent.duration._
 
 class SpanSpec extends AnyFlatSpec with Matchers with ScalaCheckDrivenPropertyChecks with ArbitraryInstances {
   val ec: TestContext = TestContext()
-  implicit val timer: Timer[IO] = ec.ioTimer
-  implicit val ctx: ContextShift[IO] = ec.ioContextShift
-
-  val sc: ScheduledExecutorService = new ScheduledThreadPoolExecutor(1)
+  implicit val ioRuntime: IORuntime = {
```
catostrophe, line 21: This actually hangs for eternity. I think we should migrate to a framework that allows writing tests without calling `unsafeRunSync()`. Will do it later. FS2 uses munit.

modules/avro-exporter/src/main/scala/io/janstenpickle/trace4cats/avro/AvroSpanExporter.scala
```diff
@@ -98,22 +98,23 @@ object AvroSpanExporter {
           .start
       )(fiber =>
         Applicative[F].unit.whileM_(for {
-          queueNonEmpty <- queue.getSize.map(_ != 0)
+          queueNonEmpty <- Sync[F].delay[Boolean](
```
catostrophe, line 101: `cats.effect.std.Queue` doesn't support peeking for now. Something needs to be done here...

modules/avro-exporter/src/main/scala/io/janstenpickle/trace4cats/avro/AvroSpanExporter.scala
```diff
@@ -151,27 +152,30 @@ object AvroSpanExporter {
...
-      queue <- Resource.liftF(InspectableQueue.bounded[F, Batch[G]](1))
-      semaphore <- Resource.liftF(Semaphore[F](Long.MaxValue))
-      socketGroup <- TCPSocketGroup(blocker)
+      avroSchema <- Resource.eval(AvroInstances.completedSpanSchema[F])
+      h <- Resource.eval(Host.fromString(host).liftTo[F](new IllegalArgumentException(s"invalid host $host")))
+      p <- Resource.eval(Port.fromInt(port).liftTo[F](new IllegalArgumentException(s"invalid port $port")))
+      address = SocketAddress(h, p)
+      queue <- Resource.eval(Queue.bounded[F, Batch[G]](1))
+      semaphore <- Resource.eval(Semaphore[F](Long.MaxValue))
+      socketGroup <- Network[F].socketGroup()
```
catostrophe, line 161: `Blocker` is not needed anymore. But threads under `socketGroup` and `datagramSocketGroup` can be configured.

catostrophe:
@janstenpickle @CremboC I would appreciate your review.

The migration itself was quite easy. But there are several decisions to question and maybe to rework. I marked them with TODOs.

For now, we're blocked by a few dependencies and tests. I couldn't manage to work tests with the new IORuntime. It hangs for some reason when used with `TestContext`.

There are also a few things to change:
1) fs2.InspectableQueue is now deprecated. cats.effect.std.Queue doesn't support getSize. We need to rewrite the code that uses it.
2) There a lot of builder around BlazeClient that now requires an EC. It used to requier a `Blocker` but now it's gone. Moreover, it is said in the docs that EC.global would fit by default, so BlazeClient doesn't block. We can use `Async[F].executionContext` which is the default EC of the Runtime. Or ask for an optional EC param and use the default EC if it's None.
3) Some exporters wrap java drivers which I suppose to be non-blocking (udp, grpc senders). We used to provide Blockers for wrapping calls to them. I kept this semantic but I encourage you to discuss whether it's really needed to block such calls. They seem to be fire-and-forget calls to non-blocking drivers.

I also changed `model` a bit. `SpanId` and `TraceId` now don't conceal access to ThreadLocal random. Now they require `Random` as a bound on F. And where we have a `Sync` in the scope I derive a `Random[F]` instance via ThreadLocal random. I hope this change is ok for you. IMO, it is more fair and explicit.

modules/avro-exporter/src/main/scala/io/janstenpickle/trace4cats/avro/AvroSpanExporter.scala
```diff
@@ -84,12 +82,16 @@ object AvroSpanExporter {
...
-      socket <- socketGroup.open()
+      avroSchema <- Resource.eval(AvroInstances.completedSpanSchema[F])
+      host <- Resource.eval(IpAddress.fromString(host).liftTo[F](new IllegalArgumentException(s"invalid host $host")))
+      port <- Resource.eval(Port.fromInt(port).liftTo[F](new IllegalArgumentException(s"invalid port $port")))
+      address = SocketAddress(host, port)
+      queue <- Resource.eval(
+        Queue.bounded[F, Batch[G]](1)
+      ) //TODO: replace with Ref of Option or Queue of Option and noneTerminate?
+      semaphore <- Resource.eval(Semaphore[F](Long.MaxValue))
+      socketGroup <- Network[F].datagramSocketGroup() //TODO: maybe we should use the global datagramSocketGroup
```
catostrophe, line 93: https://gitter.im/functional-streams-for-scala/fs2?at=603fdc6ad74bbe49e0cf8fe5

There's no reason to use manually managed socket groups

modules/newrelic-http-exporter/src/main/scala/io/janstenpickle/trace4cats/newrelic/NewRelicSpanCompleter.scala
```diff
@@ -10,27 +10,29 @@ import io.janstenpickle.trace4cats.model.TraceProcess
 import org.http4s.client.Client
 import org.http4s.client.blaze.BlazeClientBuilder
 
+import scala.concurrent.ExecutionContext
+
 object NewRelicSpanCompleter {
-  def blazeClient[F[_]: ConcurrentEffect: Timer](
-    blocker: Blocker,
+  def blazeClient[F[_]: Async](
+    ec: ExecutionContext, //TODO: keep parameter or replace with EC.global as recommended?
```
catostrophe, line 17: Here and everywhere:
https://gitter.im/http4s/http4s?at=603fd64f457d6b4a949ed0fe

There's no reason for using manually provided EC.

janstenpickle:
Thanks @catostrophe this is really good.

I think some of the changes you mention can be backported now:

- Not using `Blocker` for Async Java exporters
- Do not use manually managed socket groups for avro exporters
- Possibly switch tests to Munit?
- Remove any usage of peeking on FS2 queues
- Invert control of `Random` for span and trace IDs (I agree this is much clearer and makes it easier to test!)
- Optional parameter for an execution context for Blaze client and server

I'm happy to start knocking some of these off over the weekend!

catostrophe:
@janstenpickle 

> * Not using `Blocker` for Async Java exporters
> * Optional parameter for an execution context for Blaze client and server

I made all ECs/Blockers optional. Since now we may ask Runtime to run a task on the default working EC or the default blocker. Now API looks better: if you want more fine-grained pool tuning, you provide your own EC; if not, just use the runtime defaults.

> * Do not use manually managed socket groups for avro exporters

I looked into the implementation of `Network[F]` and found that the global socket groups are initialized lazily NOT as resources. So manually managed socket groups seem even a cleaner approach. I kept them. I see no harm in using them.

> * Possibly switch tests to Munit?

I support this.

catostrophe:
Now I'm looking at google-cloud-trace interop and I see there is Async API. I'll try to get rid of blocking calls.

modules/opentelemetry-common/src/main/scala/io/janstenpickle/trace4cats/opentelemetry/common/OpenTelemetryGrpcSpanExporter.scala
```diff
@@ -24,40 +22,32 @@ object OpenTelemetryGrpcSpanExporter {
...
-          exporter
-            .`export`(
-              batch.spans
-                .foldLeft(ListBuffer.empty[SpanData]) { (buf, span) =>
-                  buf += Trace4CatsSpanData(Trace4CatsResource(span.serviceName), span)
-                }
-                .asJavaCollection
-            )
-        )
+        Sync[F]
```
catostrophe, line 39: @janstenpickle please look at this commit carefully

I think there was a bug. `exporter.export(...)` was called eagerly without wrapping in `Sync[F].delay`
This might lead to earlier execution.

I refactored interop with `CompletableResultCode` and make a special typeclass for it (inspired by `monix.catnap.FutureLift`)

This should be backported to `master`. Maybe it is somehow related to the runtime failures I got with this exporter.

catostrophe:
All GRPC calls are async now.
All senders have an extra optional parameter ExecutionContext. It should be passed if a user wants an exporter to makes its calls in a dedicated thread pool. Most of senders are async, some are blocking - the difference is in the parameter name: `ec` or `blocker`.
If a EC is not passed, the exported works in the default runtime pool (be it the main worker pool or a blocking pool - depends on the sender).

modules/opentelemetry-common/src/main/scala/io/janstenpickle/trace4cats/opentelemetry/common/OpenTelemetryGrpcSpanExporter.scala
```diff
@@ -27,37 +28,42 @@ object OpenTelemetryGrpcSpanExporter {
...
+      for {
+        spans <- Sync[F].delay(
+          spans
+            .foldLeft(ListBuffer.empty[SpanData]) { (buf, span) =>
+              buf += Trace4CatsSpanData(Trace4CatsResource(span.serviceName), span)
+            }
+            .asJavaCollection
+        )
+        call = liftCompletableResultCode(Sync[F].delay(exporter.`export`(spans)))(ExportFailure(host, port))
+        _ <- ec.fold(call)(call.evalOn)
```
catostrophe, line 54: Ouch, this makes no sense... Will remove

modules/stackdriver-grpc-exporter/src/main/scala/io/janstenpickle/trace4cats/stackdriver/StackdriverGrpcSpanExporter.scala
```diff
@@ -130,21 +133,45 @@ object StackdriverGrpcSpanExporter {
...
+                .foldLeft(scala.collection.mutable.ListBuffer.empty[Span]) { (buf, span) =>
+                  buf += convert(span)
+                }
+                .asJava
+            )
+            .build()
         )
-      )
+        call = liftApiFuture(Sync[F].delay(client.batchWriteSpansCallable().futureCall(request)))
+        _ <- ec.fold(call)(call.evalOn)
```
catostrophe, line 169: This makes no sense. Will revert

CremboC: I know you already reverted this (I agree with reverting it), but in general I think the library should be more prescriptive rather than give lots of small choices like this. If the underlying call is thread- blocking, then we should use `Blocker`, if it's not - then don't. Giving options like this only complicates the code and makes it less maintainable.

catostrophe: There's no more Blocker in CE3. I left explicit (yet optional) ECs only where they're required by the API we interop with.

Optional because CE3 effect _knows_ much more about the runtime and has access to the main worker EC and the internal blocker EC.

catostrophe: Now you should pass EC _only_ when you really has a special reason for it.

modules/avro-exporter/src/main/scala/io/janstenpickle/trace4cats/avro/AvroSpanExporter.scala
catostrophe: No configuration needed. Default instances are ok.

modules/avro-exporter/src/main/scala/io/janstenpickle/trace4cats/avro/AvroSpanExporter.scala
```diff
@@ -54,66 +52,70 @@ object AvroSpanExporter {
...
+      socketGroup <- Network[F].datagramSocketGroup()
+      socket <- socketGroup.openDatagramSocket()
+      _ <- Resource
+        .make(
+          Stream
+            .retry(write(avroSchema, address, semaphore, queue, socket), 5.seconds, _ + 1.second, Int.MaxValue)
+            .compile
+            .drain
+            .start
+        )(fiber =>
```
catostrophe, line 102: @janstenpickle 
I changed the condition for releasing the worker fiber.

The fiber may be in 2 states:
- processing a new batch - hence it has acquired a permit for it
- waiting on `queue.take` - hence it has released the permit

So if we check the available permits and it's less than the `maxPermits`, then the fiber is working on a new batch and we should sleep. Otherwise, the fiber is semantically blocked and there are no new messages in the queue.

Strange, but the previous check contained `semaphore.count.map(_ == 0)`. This is probably a mistake because `count` returns `maxPermits - <number of acquired permits>`.

catostrophe: Btw, what was the intention behind using a semaphore and not a Ref? Multiple fibers working concurrently?
We can add an extra parameter and use it for the queue capacity and the number of permits/workers. And we can start n fibers. Does it make sense?

catostrophe:
Everything compiles now (except for a few modules that still don't have CE3-compat dependencies).
Now going to migrate tests to Munit.

I managed to fix everything that I found but there are still a few open questions.

catostrophe:
I finally managed to migrate all modules that have CE3-compatible dependencies and make tests work.

The only test that still hangs is `io.janstenpickle.trace4cats.rate.TokenBucketSpec`. The ticker machinery is somewhat unclear to me, don't know why it hangs.

janstenpickle:
> I finally managed to migrate all modules that have CE3-compatible dependencies and make tests work.
> 
> ~The only test that still hangs is `io.janstenpickle.trace4cats.rate.TokenBucketSpec`.~ All tests have been migrated!

🙌 

Thanks @catostrophe. I'm going to start backporting some of these changes to CE2 today, then we can cut a pre-CE3 release to reduce the delta in the upgrade.

modules/avro-exporter/src/main/scala/io/janstenpickle/trace4cats/avro/AvroSpanExporter.scala
janstenpickle: > We can add an extra parameter and use it for the queue capacity and the number of permits/workers. And we can start n fibers. Does it make sense?

Yeah this makes sense. It was most likely a naive implementation on my part!

modules/tail-sampling-redis-store/src/main/scala/io/janstenpickle/trace4cats/sampling/tail/redis/RedisSampleDecisionStore.scala
```diff
@@ -90,11 +89,9 @@ object RedisSampleDecisionStore {
             Applicative[F].unit
           else
             for {
-              results <- decisions.foldLeft(Applicative[F].pure(List.empty[Fiber[F, Unit]])) {
-                case (acc, (traceId, decision)) =>
-                  cmd.setEx(keyPrefix -> traceId, decision, ttl).start.flatMap(fiber => acc.map(fiber :: _))
+              _ <- decisions.toList.parTraverse_ { case (traceId, decision) =>
```
catostrophe, line 92: This is a much simpler implementation. `Parallel` for any IO performs applicative product via `both` - this is safer and prevents fiber leakage.

catostrophe: It's worth backporting

modules/avro-exporter/src/main/scala/io/janstenpickle/trace4cats/avro/AvroSpanExporter.scala
```diff
@@ -54,69 +59,81 @@ object AvroSpanExporter {
...
+      override def exportBatch(batch: Batch[G]): F[Unit] = queue.offer(batch)
     }
   }
 
-  def tcp[F[_]: Concurrent: ContextShift: Timer: Logger, G[_]: Traverse](
-    blocker: Blocker,
+  def tcp[F[_]: Async: Logger, G[_]: Traverse](
     host: String = agentHostname,
     port: Int = agentPort,
+    numFibers: Int = 1
```
catostrophe, line 128: add docs

modules/example/src/main/scala/io/janstenpickle/trace4cats/example/Fs2AdvancedExample.scala
```diff
@@ -103,11 +102,10 @@ object Fs2AdvancedExample extends IOApp {
       }
 
   override def run(args: List[String]): IO[ExitCode] =
-    (for {
-      blocker <- Blocker[IO]
-      ep <- entryPoint[IO](blocker, TraceProcess("trace4catsFS2"))
-    } yield ep)
+    entryPoint[IO](TraceProcess("trace4catsFS2"))
       .use { ep =>
+        implicit val random: Random[Kleisli[IO, Span[IO], *]] = Random.javaUtilConcurrentThreadLocalRandom
```
catostrophe, line 107: TODO: think on not passing it
switch to our own typeclass that has derviation via cats.effects.std.Random

modules/meta/src/main/scala/io/janstenpickle/trace4cats/meta/TracedSpanExporter.scala
```diff
@@ -14,50 +14,53 @@ object TracedSpanExporter {
...
-  ): StreamSpanExporter[F] = new StreamSpanExporter[F] {
-    override def exportBatch(batch: Batch[Chunk]): F[Unit] = for {
-      context <- SpanContext.root[F]
-      sample <- sampler.shouldSample(None, context.traceId, spanName, spanKind)
-      _ <- sample match {
-        case SampleDecision.Drop => underlying.exportBatch(batch)
-        case SampleDecision.Include =>
-          val (batchSize, links) = MetaTraceUtil.extractMetadata(batch.spans)
+  ): StreamSpanExporter[F] = {
+    implicit val random: Random[F] = Random.javaUtilConcurrentThreadLocalRandom
```
janstenpickle, line 24: Needs adding to type bounds/our own "random" typeclass

janstenpickle:
@catostrophe I'm happy with this if you are. I'll create a `CE3` branch to merge into

catostrophe:
@janstenpickle fine!

PR #350, opened 2021-03-12
Get rid of InspectableQueue in AvroSpanExporter

### LINKED ISSUE #340: AvroSpanExporter shouldn't check the queue for elements on release
TODO:
1. Rely on the semaphore only.
2. Fix the check for acquired permits.
3. Add an optional parameter for the number of worker fibers (under discussion).

catostrophe:
Closes #340


modules/avro-exporter/src/main/scala/io/janstenpickle/trace4cats/avro/AvroSpanExporter.scala
```diff
@@ -69,40 +75,39 @@ object AvroSpanExporter {
...
           .compile
           .drain
           .start
       )(fiber =>
-        Applicative[F].unit.whileM_(for {
-          queueNonEmpty <- queue.getSize.map(_ != 0)
-          semaphoreSet <- semaphore.count.map(_ == 0)
-          _ <- Timer[F].sleep(50.millis)
-        } yield queueNonEmpty || semaphoreSet) >> fiber.cancel
+        Timer[F].sleep(50.millis).whileM_(semaphore.available.map(_ < maxPermits)) >>
```
catostrophe, line 107: This is probably not the best strategy for releasing multiple fibers, since every single fiber waits until no work is being done at all.

modules/avro-exporter/src/main/scala/io/janstenpickle/trace4cats/avro/AvroSpanExporter.scala
```diff
@@ -112,7 +117,11 @@ object AvroSpanExporter {
     blocker: Blocker,
     host: String = agentHostname,
     port: Int = agentPort,
+    numFibers: Int = 1
```
janstenpickle, line 120: Could you modify the tests to add a case where multiple fibers are used please?

catostrophe: Yes, a bit later. How important is the order of spans being sent by Exporter? Multiple fibers will be sending batches out of the queue order.

catostrophe: @janstenpickle done
I also noticed that the existing tests were using `Batch[List]` which arb instance returned zero-sized lists (there's minSuccessful = 1 configuration). I changed it to `Batch[Chunk]` as it's used in other tests

modules/avro-exporter/src/main/scala/io/janstenpickle/trace4cats/avro/AvroSpanExporter.scala
```diff
@@ -59,7 +62,10 @@ object AvroSpanExporter {
...
```
catostrophe, line 65: Add documentation for `numFibers` with warning

janstenpickle (approved):
Pending scaladoc, this is good!

PR #365, opened 2021-03-20
Fix fluky Jaeger tests

catostrophe:
I suddenly caught a failure in tests. It turned out that Jaeger automatically converts any tag with key "ip" from Long to String (eg. 2130706433 -> '127.0.0.1').

So I filter "ip" out.


catostrophe:
This is a backport from the CE3 branch.

PR #399, opened 2021-04-01
Proper shifting after async interop in CE2

catostrophe:
I overlooked this when we were backporting things from the CE3 branch. This is mandatory in CE2: after `async` we must shift back. Otherwise, the continuation will perform in the grpc (netty) pool.

P.S. This is NOT needed in CE3, since `Async` automatically shifts back.


janstenpickle (approved):
Oops, nice catch

PR #418, opened 2021-04-07
Drop sbt 0.13 slash syntax

Ssstlis:
This PR fixes warnings in sbt like this:
```
[info] /build.sbt:18: warning: method in in trait ScopingSetting is deprecated (since 1.5.0): `in` is deprecated; migrate to slash syntax - https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#slash
  javacOptions in (Compile, compile) ++= Seq("-source", "1.8", "-target", "1.8"),
               ^
/build.sbt:48: warning: method in in trait ScopingSetting is deprecated (since 1.5.0): `in` is deprecated; migrate to slash syntax - https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#slash
  publishArtifact in Test := false
  
  ```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,7 +15,7 @@ lazy val commonSettings = Seq(
   scmInfo := Some(
     ScmInfo(url("https://github.com/janstenpickle/trace4cats"), "scm:git:git@github.com:janstenpickle/trace4cats.git")
   ),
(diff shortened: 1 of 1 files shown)


catostrophe:
Fixed in #409

PR #439, opened 2021-04-22
Derivations via magnolia

catostrophe:
Switch from `kittens` and `scalacheck-shapeless` to `magnolify` as a more performant alternative.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -155,7 +155,7 @@ lazy val model =
         Dependencies.enumeratum,
         Dependencies.enumeratumCats,
         Dependencies.commonsCodec,
-        Dependencies.kittens
+        Dependencies.magnolifyCats
       )
     )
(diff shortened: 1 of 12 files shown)


catostrophe:
@janstenpickle unfortunately I couldn't make tests work with scalacheck `Arbitrary`s derived via `magnolify-scalacheck`.
What do you think about replacing just `kittens` with `magnolify-cats`?

janstenpickle:
> What do you think about replacing just `kittens` with `magnolify-cats`?

That sounds good to me. I'm wondering we should have some static fixtures for some of these tests, as we aren't always testing the serialization.

catostrophe:
The future of `magnolia` is uncertain, so I'm closing this for now, since it's not critical and was more of an experiment.

PR #466, opened 2021-05-06
Add attribute helpers

janstenpickle:
This should help with the proposed Zipkin exporter:

- Introduces semantic conventions for attribute names, these can
  be used with exporters like Zipkin where more information is
  useful/required
- Add helpers for extracting attributes from the local system
  such as hostname, IP address, environment variables and
  system properties
- Create an optional builder for `TraceProcess` that allows
  attributes to be resolved by side affect to be added to the
  `TraceProcess` via a fluent API
- Updated Http4s client and server wrappers to capture service
  information such as local and remote service when making
  requests
- Updated Sttp client to capture request information for the
  remote service

### DIFF
--- a/modules/model/src/main/scala/io/janstenpickle/trace4cats/model/SemanticAttributeKeys.scala
+++ b/modules/model/src/main/scala/io/janstenpickle/trace4cats/model/SemanticAttributeKeys.scala
@@ -1,0 +1,25 @@
+package io.janstenpickle.trace4cats.model
+
+object SemanticAttributeKeys {
+  // local service attributes
+  private final val service = "service."
+  final val servicePort: String = service + "port"
+  final val serviceIpv4: String = service + "ipv4"
+  final val serviceIpv6: String = service + "ipv6"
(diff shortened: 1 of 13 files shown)


modules/model/src/main/scala/io/janstenpickle/trace4cats/model/SemanticAttributeKeys.scala
```diff
@@ -0,0 +1,25 @@
+package io.janstenpickle.trace4cats.model
+
+object SemanticAttributeKeys {
+  // local service attributes
+  private final val service = "service."
+  final val servicePort: String = service + "port"
+  final val serviceIpv4: String = service + "ipv4"
+  final val serviceIpv6: String = service + "ipv6"
+  final val serviceHostname: String = service + "hostname"
```
janstenpickle, line 9: I purposefully didn't prepend these with `local` as I think we should be biased towards the local system

janstenpickle:
I'm going to rebase this against master, then we can release 0.11 as the final CE2 version

Issue #467, opened 2021-05-06
Case sensitivity in b3 headers


soujiro32167:
I am trying to see a call flow from Envoy to a backend that uses trace4cats. The incoming headers include b3 headers, like this:

![image](https://user-images.githubusercontent.com/6598104/117339547-dac5da80-ae6d-11eb-8416-80970a46f02d.png)

However, in Jaeger UI, I cannot see the connection between the parent trace from Envoy and the child created by trace4cats.
It looks like [here](https://github.com/janstenpickle/trace4cats/blob/master/modules/core/src/main/scala/io/janstenpickle/trace4cats/B3ToHeaders.scala) the headers are compared directly, with case sensitivity, whereas the protocol allows lower case headers:

https://github.com/openzipkin/b3-propagation

catostrophe:
Closed via #468

PR #468, opened 2021-05-06
Closes issue #467: trace headers are now case insensitive

soujiro32167:


janstenpickle:
Thanks for this @soujiro32167! We are partially using case insensitive strings against the cats-effect 3 branch for Http4s headers, but not in the trace headers, so this is great.

JFYI, it looks like you need to run `scalafmt`.

janstenpickle:
> Done, thanks for reviewing @janstenpickle! This is a great lib. I was able to to instrument a mid-size code base: HTTP servers/clients, plus implement a version of [https://github.com/ovotech/natchez-extras/tree/master/natchez-extras-doobie](https://github.com/ovotech/natchez-extras/tree/master/natchez-extras-doobie?rgh-link-date=2021-05-07T13%3A38%3A17Z) in two weeks =]

Thanks, I'm glad you like it! We do have interop for Natchez by the way, see the [example here](https://github.com/janstenpickle/trace4cats/blob/master/docs/examples.md#natchez). It should implicitly convert between the trace4cats `Trace` typeclass and Nathez's.

catostrophe:
TODO: port to the CE3 branch

PR #477, opened 2021-05-11
Add attribute helpers

janstenpickle:
CE2 version of #466 

- Introduces semantic conventions for attribute names, these can
  be used with exporters like Zipkin where more information is
  useful/required
- Add helpers for extracting attributes from the local system
  such as hostname, IP address, environment variables and
  system properties
- Create an optional builder for `TraceProcess` that allows
  attributes to be resolved by side affect to be added to the
  `TraceProcess` via a fluent API
- Updated Http4s client and server wrappers to capture service
  information such as local and remote service when making
  requests
- Updated Sttp client to capture request information for the
  remote service

### DIFF
--- a/modules/core/src/main/scala/io/janstenpickle/trace4cats/attributes/EnvironmentAttributes.scala
+++ b/modules/core/src/main/scala/io/janstenpickle/trace4cats/attributes/EnvironmentAttributes.scala
@@ -1,0 +1,28 @@
+package io.janstenpickle.trace4cats.attributes
+
+import cats.effect.Sync
+import cats.syntax.functor._
+import io.janstenpickle.trace4cats.model.AttributeValue
+import io.janstenpickle.trace4cats.model.AttributeValue.StringValue
+
+import scala.jdk.CollectionConverters._
+
+object EnvironmentAttributes {
+  def apply[F[_]: Sync](): F[Map[String, AttributeValue]] = filterKeys(_ => true)
+
+  def excludeKeys[F[_]: Sync](excludeKeys: Set[String]): F[Map[String, AttributeValue]] = filterKeys(
+    !excludeKeys.contains(_)
+  )
+
+  def filterKeys[F[_]: Sync](filterKeys: String => Boolean): F[Map[String, AttributeValue]] =
+    Sync[F]
+      .delay(System.getenv())
+      .map(_.asScala.toMap.flatMap { case (k, v) =>
+        if (filterKeys(k)) Some(k -> StringValue(v)) else None
+      })
+
+  def includeKeys[F[_]: Sync](includeKeys: Set[String]): F[Map[String, AttributeValue]] =
+    Sync[F].delay(System.getenv()).map { env =>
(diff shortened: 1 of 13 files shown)


modules/http4s-client/src/main/scala/io/janstenpickle/trace4cats/http4s/client/Http4sClientRequest.scala
```diff
@@ -0,0 +1,17 @@
...
+
+import io.janstenpickle.trace4cats.http4s.common.Request_
+import io.janstenpickle.trace4cats.model.AttributeValue.{LongValue, StringValue}
+import io.janstenpickle.trace4cats.model.{AttributeValue, SemanticAttributeKeys}
+import org.http4s.Uri
+
+object Http4sClientRequest {
+  def toAttributes(req: Request_): Map[String, AttributeValue] =
+    req.uri.host.map {
+      case address @ Uri.Ipv4Address(_, _, _, _) =>
```
catostrophe, line 11: why not `case address: Uri.Ipv4Address => `?

janstenpickle: Because I'm using an extractor for `host` below, and it's not good practice to mix different pattern matching styles.

modules/core/src/main/scala/io/janstenpickle/trace4cats/attributes/EnvironmentAttributes.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import cats.effect.Sync
+import cats.syntax.functor._
+import io.janstenpickle.trace4cats.model.AttributeValue
+import io.janstenpickle.trace4cats.model.AttributeValue.StringValue
+
+import scala.jdk.CollectionConverters._
+
+object EnvironmentAttributes {
+  def apply[F[_]: Sync](): F[Map[String, AttributeValue]] = filterKeys(_ => true)
```
catostrophe, line 11: are empty arg parentheses `()` needed?

modules/core/src/main/scala/io/janstenpickle/trace4cats/attributes/SystemPropertyAttributes.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import cats.effect.Sync
+import cats.syntax.functor._
+import io.janstenpickle.trace4cats.model.AttributeValue
+import io.janstenpickle.trace4cats.model.AttributeValue.StringValue
+
+import scala.jdk.CollectionConverters._
+
+object SystemPropertyAttributes {
+  def apply[F[_]: Sync](): F[Map[String, AttributeValue]] = filterKeys(_ => true)
```
catostrophe, line 11: why are empty arg parentheses () needed?

catostrophe (approved):

catostrophe:
@janstenpickle I just noticed there's no sttp-client2 support.

PR #497, opened 2021-05-14
Stop releasing after mergify[bot] PRs

catostrophe:


catostrophe:
This one seems to be the right github.actor.

PR #525, opened 2021-05-21
RateSpanSampler should respect `rootSpansOnly` (backport to CE2)

catostrophe:
Backported from the master (CE3) branch.


catostrophe:
@janstenpickle did you want to backport anything else or just this?

janstenpickle:
Thanks for this @catostrophe. I think this is enough, it's more of a fix than a feature and I think we should only be targeting new features at CE3 now.

janstenpickle (approved):

PR #528, opened 2021-05-23
Fix flaky Jaeger tests

catostrophe:
Jaeger automatically transforms some tags, so they can't be found in response in tests.


modules/zipkin-http-exporter/src/test/scala/io/janstenpickle/trace4cats/zipkin/package.scala
```diff
@@ -23,4 +23,6 @@ package object zipkin {
         case (k, AttributeValue.BooleanValue(value)) if k == "error" => JaegerTag.BoolTag(k, value.value)
         case (k, v) => JaegerTag.StringTag(k, v.show)
       }
+
+  val excludedTagKeys: Set[String] = Set("lc")
```
catostrophe, line 27: This automatically transforms to "component" 🤷

catostrophe: And only for Zipkin

PR #530, opened 2021-05-24
Add hotswappable completer & exporter implementations

janstenpickle:
Gives `SpanCompleter` and `SpanExporter` the same treatment as `SpanSampler`.


modules/exporter-common/src/main/scala/io/janstenpickle/trace4cats/export/ConfiguredHotswapSpanExporter.scala
```diff
@@ -0,0 +1,23 @@
...
+    initialConfig: A,
+    makeExporter: A => Resource[F, SpanExporter[F, G]]
+  ): Resource[F, ConfiguredHotswapSpanExporter[F, G, A]] =
+    HotswapSpanExporter[F, G, A](initialConfig, makeExporter(initialConfig)).map { underlying =>
+      new ConfiguredHotswapSpanExporter[F, G, A] {
+        override def updateConfig(config: A): F[Boolean] = underlying.update(config, makeExporter(config))
+        override def exportBatch(batch: Batch[G]): F[Unit] = underlying.exportBatch(batch)
+      }
+    }
+}
```
janstenpickle, line 23: This implementation pattern would probably be useful everywhere, I'm wondering if we make it the default for everything so that `A` and the underlying resources don't get out of sync when someone tries to use a low level implementation.

modules/exporter-common/src/main/scala/io/janstenpickle/trace4cats/export/ExportRetryConfig.scala
```diff
@@ -1,9 +1,18 @@
...
 import scala.concurrent.duration._
 
 case class ExportRetryConfig(
   delay: FiniteDuration = 500.millis,
   nextDelay: FiniteDuration => FiniteDuration = _ + 100.millis,
   maxAttempts: Int = 100
 )
+
+object ExportRetryConfig {
+  implicit val nexDelayEq: Eq[FiniteDuration => FiniteDuration] = Eq.fromUniversalEquals
```
catostrophe, line 15: Wait. You can't compare functions.

catostrophe: and there's a typo in the name `nexDelayEq`

catostrophe: I think this should work if one makes a new instance of config as `oldConfig.copy(delay = newDelay)`

catostrophe: But in general, we can't compare functions. So we may create a new resource even if it's not needed. Which is probably not so bad (compared to the opposite)

janstenpickle: Ah, I've got an idea for this

modules/exporter-common/src/main/scala/io/janstenpickle/trace4cats/export/ExportRetryConfig.scala
```diff
@@ -1,9 +1,17 @@
...
 
 case class ExportRetryConfig(
   delay: FiniteDuration = 500.millis,
   nextDelay: FiniteDuration => FiniteDuration = _ + 100.millis,
   maxAttempts: Int = 100
 )
+
+object ExportRetryConfig {
+  implicit val eq: Eq[ExportRetryConfig] = Eq.by { config =>
+    (config.delay, config.nextDelay(config.delay), config.maxAttempts)
```
catostrophe, line 15: Hm... This is riskier, I think. This instance may lead to false `eqv`. And the new instance won't be created when needed, e.g. when only the function has changed but its `next` coincides with the `next` of the previous function.

catostrophe: How can we compare 2 functions? Maybe by probing more than 2 points? But how many?

catostrophe: What are the downsides of swapping resources when the config is not changed?
What are the downsides of NOT swapping when needed?

catostrophe: Let's make ExportRetryConfig a sealed trait of 3 cases:
- Constant
- Exponential
- Fibonacci

modules/exporter-common/src/main/scala/io/janstenpickle/trace4cats/export/QueuedSpanCompleter.scala
```diff
@@ -38,7 +39,11 @@ object QueuedSpanCompleter {
               .retry(
                 exporter.exportBatch(batch),
                 delay = config.retryConfig.delay,
-                nextDelay = config.retryConfig.nextDelay,
+                nextDelay = t =>
+                  config.retryConfig.nextDelay match {
```
catostrophe, line 43: I think this expression should be a function placed within NextDelay

modules/exporter-common/src/main/scala/io/janstenpickle/trace4cats/export/ConfiguredHotswapSpanExporter.scala
catostrophe: This is a real boilerplate piece of code. And this may be abstracted over the resource.

modules/hotswap/src/main/scala/io/janstenpickle/trace4cats/hotswap/HotswapConstructor.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import cats.Applicative
+import cats.effect.kernel.{Resource, Temporal}
+import cats.kernel.Eq
+import cats.syntax.applicative._
+import cats.syntax.eq._
+import cats.syntax.functor._
+import cats.syntax.flatMap._
+
+trait HotswapConstructor[F[_], A, B] {
```
catostrophe, line 11: Can we give A, B better names?

modules/hotswap/src/main/scala/io/janstenpickle/trace4cats/hotswap/HotswapConstructor.scala
```diff
@@ -0,0 +1,36 @@
...
+  def swap(a: A): F[Boolean]
+  def get: Resource[F, (A, B)]
+  def getA: F[A]
```
catostrophe, line 14: I don't like `getA`, `getB` names. Because their return types are different and there is no hint of what they mean.

catostrophe (changes requested):

modules/exporter-common/src/main/scala/io/janstenpickle/trace4cats/export/QueuedSpanCompleter.scala
```diff
@@ -38,7 +38,7 @@ object QueuedSpanCompleter {
               .retry(
                 exporter.exportBatch(batch),
                 delay = config.retryConfig.delay,
-                nextDelay = config.retryConfig.nextDelay,
+                nextDelay = t => config.retryConfig.nextDelay.calc(t),
```
catostrophe, line 41: `nextDelay = config.retryConfig.nextDelay.calc` should work

build.sbt
```diff
@@ -244,6 +245,15 @@ lazy val base =
     .settings(publishSettings)
     .settings(name := "trace4cats-base", libraryDependencies ++= Dependencies.test.map(_ % Test))
 
+lazy val hotswap =
+  (project in file("modules/hotswap"))
+    .settings(publishSettings)
+    .settings(
+      name := "trace4cats-hotswap",
+      libraryDependencies ++= Seq(Dependencies.catsEffect),
```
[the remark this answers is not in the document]
janstenpickle: This actually needs to be `std`, I'll add the dependency and update

modules/dynamic-sampling-config/src/main/scala/io/janstenpickle/trace4cats/sampling/dynamic/config/ConfiguredHotSwapSpanSampler.scala
```diff
@@ -11,15 +11,14 @@ trait ConfiguredHotSwapSpanSampler[F[_]] extends SpanSampler[F] {
 }
 
 object ConfiguredHotSwapSpanSampler {
-  def create[F[_]: Temporal](initialConfig: SamplerConfig): Resource[F, ConfiguredHotSwapSpanSampler[F]] =
-    HotSwapSpanSampler
-      .create(initialConfig, SamplerUtil.makeSampler(initialConfig))
+  def apply[F[_]: Temporal](initialConfig: SamplerConfig): Resource[F, ConfiguredHotSwapSpanSampler[F]] =
+    HotSwapSpanSampler(initialConfig, (c: SamplerConfig) => SamplerUtil.makeSampler[F](c))
```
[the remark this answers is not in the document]
janstenpickle: Previous versions of this were being a bit flaky, it seems to work now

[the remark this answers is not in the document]
catostrophe: this is better: `HotSwapSpanSampler(initialConfig)(SamplerUtil.makeSampler[F])`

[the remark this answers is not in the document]
janstenpickle: I'm not a big fan of separate arguments for things like this, especially when both the params are closely related

modules/hotswap/src/main/scala/io/janstenpickle/trace4cats/hotswap/HotswapRefConstructor.scala
```diff
@@ -0,0 +1,46 @@
...
+  /** Swap `R` using `I` using some function of `I` to [[cats.effect.kernel.Resource]] of `R`, provided at construction
+    * time
+    */
+  def swapWith(next: I): F[Unit] = swap(func(next).map(next -> _))
+
+  /** Access the current input value `I` on its own
+    */
+  def accessI: Resource[F, I] = access.map(_._1)
+
+  /** Access the current input value `R` on its own
```
catostrophe, line 21: wrong description

catostrophe (approved):

janstenpickle:
Just need to wait for maven central to update with the new [hotswap library](https://github.com/janstenpickle/hotswap-ref)

Issue #229, opened 2021-01-22
Cross-compiling T4C with Scala 3


catostrophe:
Below is the list of issues to be resolved to make it compile with Scala 3.

catostrophe:
- `better-monadic-for` plugin is not supported and will probably never be
- some libs are not published for Scala 3 at all and will probably never be - we should use their 2.13 versions in the dotty compat mode
- some libs are published for Scala 3, but not the latest version - we should wait for their releases rather than build with Scala 3.0.0-M2
- some code hygiene is needed
- more to be discovered

ybasket:
Scala 3 final is out, so it would be cool if trace4cats started to cross-build. I can offer some help on that as well :)

Updates to the list above:
- log4cats, vulcan and fs2-kafka are published for Scala 3
- magnolia has a Scala 3 version that's slightly different from its discontinued Scala 2 predecessor => probably requires specific sources (or usage of shapeless-deriving on Scala 3)
- `better-monadic-for` won't have a Scala 3 version, but some of the syntax is available under `-source:future`
- builds with mixed 2.13 and 3.0 dependencies have proven too be fragile so far (esp if they depend on the same lib), so ideally no mix necessary

catostrophe:
@ybasket 
`vulcan-generic`, `vulcan-enumeratum` are not published

I tried to migrate to `mangolia`, but the derived instances for tests didn't work well. So we'll stick to `kittens`, which should be released soon, since `shapless3` is out.

catostrophe:
@ybasket feel free to make a PR if you can.
The problem is we use `enumeratum`. We can try to get rid of it. I don't think we use `findValues`. But we rely on some integrations and derivations.

It would be great to start releasing at least some of the modules.

catostrophe:
Kittens migrated to Scala3/Shapeless3 and will be released soon

ybasket:
@catostrophe Just opened https://github.com/janstenpickle/trace4cats/pull/534 for `enumeratum` and `better-monadic-for`.

I also experimented with just adding Scala 3 to the build (see my branch here: https://github.com/ybasket/trace4cats/tree/scala-3). Some modules compile just fine, so far the code itself seems relatively unproblematic. It's the dependencies that will cause headache, `vulcan-generic` and even worse, `circe-generic-extras`. 

[...]

PR #534, opened 2021-05-25
Scala 3 preparations: enumeratum and better-monadic-for

ybasket:
This PR removes enumeratum as a dependency (wasn't used too much anyway) and rewrites for-comprehensions in a way that the code compiles also without the `better-monadic-for` compiler plugin, as it will be the case on Scala 3 (there's `source:future`, but the rewrites are IMHO mostly not uglier to read than the original versions, so easier that way).

Contributes to resolving #229 in baby steps.

### DIFF
--- a/modules/agent-kafka/src/main/scala-2.13/io/janstenpickle/trace4cats/agent/kafka/AgentKafka.scala
+++ b/modules/agent-kafka/src/main/scala-2.13/io/janstenpickle/trace4cats/agent/kafka/AgentKafka.scala
@@ -46,21 +46,26 @@ object AgentKafka
     bufferSize: Int,
     trace: Boolean,
     traceRate: Option[Double]
-  ): IO[ExitCode] = (for {
-    implicit0(logger: Logger[IO]) <- Resource.eval(Slf4jLogger.create[IO])
-
-    kafkaExporter <- AvroKafkaSpanExporter[IO, Chunk](kafkaBootstrapServers, kafkaTopic)
-  } yield CommonAgent.run[IO](
-    port,
-    bufferSize,
-    "Avro Kafka",
-    Map[String, AttributeValue](
-      "bootstrap.servers" -> AttributeValue.StringList(kafkaBootstrapServers),
-      "topic" -> kafkaTopic
-    ),
-    kafkaExporter,
-    s"Kafka topic '$kafkaTopic'",
-    trace,
-    traceRate
-  )).use(identity)
+  ): IO[ExitCode] =
+    Resource
+      .eval(Slf4jLogger.create[IO])
+      .flatMap { implicit logger: Logger[IO] =>
+        AvroKafkaSpanExporter[IO, Chunk](kafkaBootstrapServers, kafkaTopic)
+          .map(kafkaExporter =>
+            CommonAgent.run[IO](
+              port,
+              bufferSize,
+              "Avro Kafka",
+              Map[String, AttributeValue](
+                "bootstrap.servers" -> AttributeValue.StringList(kafkaBootstrapServers),
+                "topic" -> kafkaTopic
+              ),
+              kafkaExporter,
+              s"Kafka topic '$kafkaTopic'",
+              trace,
+              traceRate
+            )
+          )
+      }
+      .use(identity)
 }
--- a/modules/model/src/main/scala/io/janstenpickle/trace4cats/model/SpanStatus.scala
+++ b/modules/model/src/main/scala/io/janstenpickle/trace4cats/model/SpanStatus.scala
@@ -76,6 +76,12 @@ object SpanStatus {
     override lazy val isOk: Boolean = false
   }
 
-  implicit val eq: Eq[SpanStatus] = Eq.by(_.canonicalCode)
-  implicit val show: Show[SpanStatus] = Show.show(_.entryName)
+  implicit val eq: Eq[SpanStatus] = Eq.by {
+    case s @ Internal(msg) => (s.canonicalCode, Some(msg))
+    case s => (s.canonicalCode, None)
+  }
+  implicit val show: Show[SpanStatus] = Show.show {
+    case s @ Internal(msg) => s"${s.entryName}($msg)"
+    case s => s.entryName
+  }
 }
--- a/modules/model/src/main/scala/io/janstenpickle/trace4cats/model/SpanStatus.scala
+++ b/modules/model/src/main/scala/io/janstenpickle/trace4cats/model/SpanStatus.scala
@@ -1,81 +1,81 @@
 package io.janstenpickle.trace4cats.model
 
-import enumeratum.EnumEntry.Camelcase
-import enumeratum._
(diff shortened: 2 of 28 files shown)


catostrophe:
@ybasket thanks, will look tmrw

janstenpickle:
@ybasket thanks very much for this, I'll try and take a look later today as well

modules/model/src/main/scala/io/janstenpickle/trace4cats/model/SpanStatus.scala
```diff
@@ -1,81 +1,81 @@
...
     override lazy val canonicalCode: Int = 15
     override lazy val isOk: Boolean = false
   }
-  case object Unauthenticated extends SpanStatus with Camelcase {
+  case object Unauthenticated extends SpanStatus("Unauthenticated") {
     override lazy val canonicalCode: Int = 16
     override lazy val isOk: Boolean = false
   }
+
+  implicit val eq: Eq[SpanStatus] = Eq.fromUniversalEquals
```
janstenpickle, line 79: Just a quick note on this; now we're not using enumeratum I think it might be better to check equality on `canonicalCode`

modules/agent-kafka/src/main/scala-2.13/io/janstenpickle/trace4cats/agent/kafka/AgentKafka.scala
```diff
@@ -46,21 +46,26 @@ object AgentKafka
...
-    ),
-    kafkaExporter,
-    s"Kafka topic '$kafkaTopic'",
-    trace,
-    traceRate
-  )).use(identity)
+  ): IO[ExitCode] =
+    Resource
+      .eval(Slf4jLogger.create[IO])
+      .flatMap { implicit logger: Logger[IO] =>
```
catostrophe, line 52: This pattern is so repetitive, I even would like to have syntax extensions for the `Slf4jLogger` object:
```scala
Slf4jLogger.provide[F](implicit logger => something)  // something: F[A]
Slf4jLogger.provideResource[F](implicit logger => something)  // something: Resource[F, A]
```

I have something similar in my private codebase.

@janstenpickle wdyt?

catostrophe: But maybe it's not worth it...

modules/model/src/main/scala/io/janstenpickle/trace4cats/model/SpanStatus.scala
catostrophe: Btw, are 2 internal statuses with different messages still equal?

catostrophe: What should `Internal("msg").show` produce?

modules/model/src/main/scala/io/janstenpickle/trace4cats/model/SpanStatus.scala
```diff
@@ -1,81 +1,81 @@
...
     override lazy val isOk: Boolean = false
   }
-  case object Unauthenticated extends SpanStatus with Camelcase {
+  case object Unauthenticated extends SpanStatus("Unauthenticated") {
     override lazy val canonicalCode: Int = 16
     override lazy val isOk: Boolean = false
   }
+
+  implicit val eq: Eq[SpanStatus] = Eq.by(_.canonicalCode)
+  implicit val show: Show[SpanStatus] = Show.show(_.entryName)
```
catostrophe, line 80:
```suggestion
  implicit val eq: Eq[SpanStatus] = Eq.by {
    case s @ Internal(msg) => (s.canonicalCode, Some(msg))
    case s => (s.canonicalCode, None)
  }
  implicit val show: Show[SpanStatus] = Show.show {
    case s @ Internal(msg) => s"${s.entryName}($msg)"
    case s => s.entryName
  }
```

janstenpickle: this looks good 👍

modules/agent-kafka/src/main/scala-2.13/io/janstenpickle/trace4cats/agent/kafka/AgentKafka.scala
janstenpickle: Maybe something that could be contributed to log4cats directly?

catostrophe: nevermind. now it's a one-off change. further changes are not as often.

modules/model/src/main/scala/io/janstenpickle/trace4cats/model/SpanStatus.scala
catostrophe: see a suggestion below

modules/model/src/main/scala/io/janstenpickle/trace4cats/model/SpanStatus.scala
ybasket: Indeed. Committed the suggestion just now 👍

catostrophe:
@ybasket could you please also remove bm4 plugin from `build.sbt`?

ybasket:
> @ybasket could you please also remove bm4 plugin from build.sbt?

I for sure could (and will do if you want), but I suggest to leave it in. It's easy (and for `kind-projector` necessary) to have it only on Scala 2 builds and there it still provides a benefit by eliminating the final `map(r => r)` Scala 2 generates on for comprehensions which just yield their result. I've done this on several Scala 3 ports by now, encountered no issues.

janstenpickle:
> > @ybasket could you please also remove bm4 plugin from build.sbt?
> 
> I for sure could (and will do if you want), but I suggest to leave it in. It's easy (and for `kind-projector` necessary) to have it only on Scala 2 builds and there it still provides a benefit by eliminating the final `map(r => r)` Scala 2 generates on for comprehensions which just yield their result. I've done this on several Scala 3 ports by now, encountered no issues.

I wasn't aware that it's required by `kind-projector`. I'm happy to leave it in for Scala 2 purely on basis the final map elimination. My only (very slight) concern is that code which relies on it could creep back in before we start Scala 3 crossbuilds. If no one else has any other objections, I'm happy to leave it in.

ybasket:
> I wasn't aware that it's required by kind-projector

It's not, sorry for the vague explanation, both plugins are independent. 

What is needed for `kind-projector` is inclusion based on Scala version (if Scala 2, then add it, if Scala 3, then use the compiler flag `-Ykind-projector`). So this little ugliness is already needed and then it doesn't make much difference whether you add one or two compiler plugins.

catostrophe:
I agree, leave it as is.

catostrophe (approved):

janstenpickle (approved):

PR #572, opened 2021-06-28
Support Cloud Trace headers

desbo:
Adds a `ToHeaders` implementation for Google Cloud Trace headers.


modules/core/src/main/scala/io/janstenpickle/trace4cats/GoogleCloudTraceToHeaders.scala
```diff
@@ -0,0 +1,60 @@
+package io.janstenpickle.trace4cats
+
+import cats.implicits._
```
janstenpickle, line 3: Could you change this to the ala-carte import please? `cats.syntax.either._`

janstenpickle (commented):
Looks good, I have a couple of nitpiks though.

I'm just thinking that this might better belong in the Cloud Trace repo: https://github.com/trace4cats/trace4cats-cloudtrace. What do you think @catostrophe?

catostrophe:
> Looks good, I have a couple of nitpiks though.
> 
> I'm just thinking that this might better belong in the Cloud Trace repo: https://github.com/trace4cats/trace4cats-cloudtrace. What do you think @catostrophe?

@janstenpickle this PR doesn't add new dependencies, it looks consistent with B3, w3c, and envoy headers. `ToHeaders.all` doesn't need additions. I think it's fine to go to the main repo.

OTOH: what if it depended on some extra google lib? Then I would put it in that cloudtrace repo.

janstenpickle:
sorry for the delay on this, I agree that because it doesn't add any extra dependencies it can probably live in here. I'm wondering whether it might not be a great idea including it in the `all`, which we use by default everywhere, and I think we should probably rename `all` in that case. The others are cross-platform standards (well maybe not Envoy), but this is specific to Google's cloud, so someone using `all` on AWS wouldn't get any benefit.

The only concern I had was about discovery, we have a separate repo for Cloud Trace, so it _might_ make sense for it to go in there, but I'm not opposed to including it here.

desbo:
One effect of moving `ToHeaders` implementations to dedicated modules would be that users need to know these exist, and choose which ones to include when creating an `EntryPoint`. I think it's quite nice that `all` is a default in `EntryPoint` and works with whatever headers the library supports without requiring any extra input from the user.

janstenpickle:
@catostrophe and I were discussing this offline and we're fine with not moving to a dedicated module, but having headers that don't have an associated open standard as part of the default header collection doesn't feel right (this prompted us to rethink the inclusion of Envoy headers).

While it doesn't completely solve discovery, what I propose we do is have a `standard` collection all the headers defined by open standards and maintain the `all` collection with all header implementations. We would then use `standard` as the default everywhere and update documentation with pointers to `all` if people want to use non-standard headers as well.

So `standard` would have `w3c |+| b3 |+| b3Single` while `all` would be `standard |+| envoy |+| googleCloudTrace`.

catostrophe (approved):

PR #587, opened 2021-07-24
Add Scala 3 build

ybasket:
Replaces kittens and scalacheck-shapeless derivations with manually spelt out instances to allow for Scala 3 cross-compilation. Both are not likely to release Scala 3-compatible versions any time soon, but their usage in trace4cats is simple and they could be re-added any time without big effort (I left the dependency declarations as comments). Meant as a proposal to bring trace4cats to Scala 3 (which would be very cool, it's the last blocker at $WORK for me), but I would understand as well if you preferred to wait.

With removing kittens, the transitive dependency on alleycats is lost. No big problem though, it's only used once in the Jaeger integration and as Jaeger is ready to compile to Scala 3 (I tried it) as soon as this PR is merged, it can then be resolved right away.


build.sbt
```diff
@@ -64,15 +64,15 @@ lazy val model =
...
+        //Dependencies.kittens, // TODO re-add once compatible with Scala 3
         Dependencies.caseInsensitive
       )
     )
 
 lazy val testkit = (project in file("modules/testkit"))
   .settings(publishSettings)
   .settings(name := "trace4cats-testkit", libraryDependencies ++= Dependencies.test ++ Seq(Dependencies.fs2))
-  .dependsOn(model)
+  .dependsOn(model % "compile->compile,test")
```
ybasket, line 75: Needed to get the `Arbitrary` instances for `AttributeValue`.

modules/base/src/test/scala/io/janstenpickle/trace4cats/base/context/KleisliInstanceSummonTest.scala
```diff
@@ -1,6 +1,7 @@
 package io.janstenpickle.trace4cats.base.context
 
 import cats.data.Kleisli
+import cats.catsInstancesForId
```
ybasket, line 4: These are necessary for Scala 3 as the implicit resolution changed.

modules/core/src/main/scala/io/janstenpickle/trace4cats/Span.scala
```diff
@@ -156,6 +156,8 @@ object Span {
 
   def noop[F[_]: Applicative]: Resource[F, Span[F]] = Resource.pure[F, Span[F]](NoopSpan[F](SpanContext.invalid))
 
+  def noopInstance[F[_]: Applicative]: Span[F] = NoopSpan[F](SpanContext.invalid)
```
ybasket, line 159: Before 0.12, one could create `NoopSpan` directly, very useful for tests. Now that the constructor is private, I smuggled this little helper in.

modules/meta/src/test/scala/io/janstenpickle/trace4cats/meta/TracedSpanExporterSpec.scala
```diff
@@ -101,7 +101,11 @@ class TracedSpanExporterSpec
           _ <- Resource.eval(tracedExporter.exportBatch(batch))
 
           allSpans <- Resource.eval(
-            Stream.fromQueueUnterminated(queue).take(batch.spans.size + expectedMetaSpans).compile.to(Chunk)
+            Stream
+              .fromQueueUnterminated(queue)
+              .take(batch.spans.size + expectedMetaSpans)
+              .compile
+              .to(Chunk): IO[Chunk[CompletedSpan]]
```
ybasket, line 108: For some weird reason, the Scala 3 compiler can't figure that out by itself, gives up on `collector.Out`.

modules/model/src/test/scala/io/janstenpickle/trace4cats/model/ArbitraryAttributeValues.scala
```diff
@@ -0,0 +1,52 @@
+package io.janstenpickle.trace4cats.model
```
ybasket, line 1: All changes due to `kittens` are above, all due to `scalacheck-shapeless` below.

catostrophe:
@ybasket thanks for your work. I am ok with this move, but we need to make sure the manually derived instances are coincide with those derived via kittens and shapeless.

I once tried to migrate to magnolify and a lot of tests started failing. 

Also, a Show instance derived via toString is different from one derived with kittens.

ybasket:
> @ybasket thanks for your work. I am ok with this move, but we need to make sure the manually derived instances are coincide with those derived via kittens and shapeless.
> 
> I once tried to migrate to magnolify and a lot of tests started failing.
> 
> Also, a Show instance derived via toString is different from one derived with kittens.

@catostrophe The point about `Show` was a good one, I just fixed the one instance I defined (format checked on Scastie against the kittens version).

All tests pass. What else would you envision for ensuring the instances are correct? 

At least what kittens derives for `Eq` should semantically be equal to the `Eq` instances of tuples as I use them in this PR, both go field-wise `&&` with `Eq` instances in scope for the field types. The only risk for the future is that one might add fields without updating the `Eq` instance, but as the model is pretty stable and the code well structured, I see this risk as not too big. One could alternatively try to derive the `Eq` instances with kittens on Scala 2 and with shapeless-3-deriving on Scala 3, but that adds more complexity (and dependencies) as you have to proxy the derivation in Scala version-specific sources.

catostrophe:
@ybasket 
> What else would you envision for ensuring the instances are correct?

I think some issues in `Arbitrary` instances may appear when you try to cross-build other modules.

We are about to cut 0.12.0 stable as soon as http4s 0.23.0 is released. This may happen in a week. I would like to postpone the release if we can add Scala 3 support. But we should know for sure we don't have blockers. Have you tried to cross-build the other repos?

ybasket:
> We are about to cut 0.12.0 stable as soon as http4s 0.23.0 is released. This may happen in a week. I would like to postpone the release if we can add Scala 3 support. But we should know for sure we don't have blockers. Have you tried to cross-build the other repos?

@catostrophe I tried quickly with the Jaeger module (which isn't too nice because you also need the integration test module first – why are they split btw?), there it seemed to work. I wanted feedback on the approach first before migrating everything, so it was only exploratory. Just tried with http4s, there the `compile->test` dependency projection seemed problematic (but that's a guess, didn't dig deep enough to be sure). Nothing that can't be solved, but a sign of risk when rushing things too quickly.

A week sounds tight to get this into a stable release state, especially given the amount of repos to consider. I definitely can help migrating some of the modules, but my time is limited this week and this PR is only the start. Would it be an option to hold back this PR, release `0.12` (I guess people are waiting for a stable CE 3 release as well) and then start a 0.13 series with Scala 3 support? I know cutting releases and juggling branches is a burden for maintainers, but maybe that's the most relaxed and safe option?

catostrophe:
Jaeger-integration-test is shared between 2 different repos, that's why it's extracted

janstenpickle:
Thanks for all this work @ybasket!

@catostrophe and I spoke about this yesterday and we think this should be done before we release `0.12.0`. Some of the libraries may be incompatible with Scala 3 for, as yet, unknown reasons, but we'll just not release those for Scala 3 (this is one of the reasons for splitting up the project anyway!)

I'll have a pass over this, but I'm happy to go ahead with the migration.

ybasket:
Alright, 0.12 it is then!

Please don't merge just yet though, I think I have to extract the `Arbitrary[AttributeValue]` into a module first as the `compile->compile,test` dependency projection doesn't seem to work smoothly when using modules of this repo in others. Should have the chance to try that later today.

ybasket:
> Please don't merge just yet though, I think I have to extract the `Arbitrary[AttributeValue]` into a module first as the `compile->compile,test` dependency projection doesn't seem to work smoothly when using modules of this repo in others. Should have the chance to try that later today.

That seems resolved now, I opened a PR on http4s which I tested with a locally published version, all tests passed. Will look at Jaeger and maybe sttp next.

janstenpickle (approved):
LGTM 👍 just a shame none of the generic libraries work at the moment

ybasket:
🎉 How do you intend to approach the PRs depending on this one? Shall I use the snapshot build (`0.12.0-RC2+17-d73c7ff3`) to make CI pass or leave them at `0.12.0-RC2` and let the builds fail? We could also agree on a version like `0.12.0-RC3-SNAPSHOT` and use that everywhere, including repos others depend on themselves (like `http4s` or `jaeger-integration-test`).

catostrophe:
@ybasket I think, you may use 0.12.0-RC2+17-d73c7ff3 in your PRs

PR #597, opened 2021-08-05
Fix Arbitrary instances in testkit

### LINKED ISSUE #596: New Arbitrary generators cause flaky tests in other repos
New instances in ArbitraryAttributeValues use an unrefined Double generator. This causes very often failures in tests in `trace4cats-opentelemetry`.

I suggest changing the double generator to avoid floating-point positions that may cause formatting of the result in the scientific format, i.e. 0.00000123 -> 1.23E-6.

Do we lose anything with this change?

cc @ybasket @janstenpickle

catostrophe:
Closes #596.

cc @ybasket

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,10 @@ updates:
       - "catostrophe"
     labels:
       - "dependency-update"
+    ignore:
+      # managed by sbt-github-actions
+      - dependency-name: "olafurpg/setup-scala"
+      - dependency-name: "actions/checkout"
+      - dependency-name: "actions/cache"
+      - dependency-name: "actions/upload-artifact"
+      - dependency-name: "actions/download-artifact"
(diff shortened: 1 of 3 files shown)


ybasket (commented):
Code looks good to me.

I only wonder what happens if a user sets an attribute to a value that has scientific representation? It's seems like a hidden type constraint to me. Is opentelemetry really not able to handle the full range of doubles?

catostrophe:
> I only wonder what happens if a user sets an attribute to a value that has scientific representation? It's seems like a hidden type constraint to me. Is opentelemetry really not able to handle the full range of doubles?

Otel driver can handle any doubles when sending spans.

Actually, the issue is in the way we compare the original batch sent to Jaeger and the result we retrieve from Jaeger via HTTP Trace API, which is internal in fact.

Jaeger returns lists as strings. And when it converts a list of doubles into a string it formats them unpredictably.
So when we compare the result we compare strings and fail. Adjusting doubles can help.

catostrophe:
Although, I'm not sure why we don't encode `AttributeList`s as `common.proto.ArrayValue`s.

I tried to pass it to Jaeger via HTTP API and it responded with 400:
json: cannot unmarshal array into Go value of type map[string]json.RawMessage

catostrophe:
Since the issue is related only to double lists, I adjusted only one double gen used for their generation.

janstenpickle (approved):

PR #608, opened 2021-09-01
Add instances for Lift[F, F] and Provide[Kleisli[F, R, *], Kleisli[F, R, *], R]

bplommer:
Needed for https://github.com/trace4cats/trace4cats-http4s/pull/21

### DIFF
--- a/modules/base/src/main/scala/io/janstenpickle/trace4cats/base/context/ContextRoot.scala
+++ b/modules/base/src/main/scala/io/janstenpickle/trace4cats/base/context/ContextRoot.scala
@@ -29,14 +29,13 @@ private[context] trait ContextRootInstancesLowPriority {
     def lift[A](fa: F[A]): F[A] = fa
   }
 
-  implicit def kleisliIdProvide[F[_]: Monad, R]: Provide[Kleisli[F, R, *], Kleisli[F, R, *], R] =
-    new Provide[Kleisli[F, R, *], Kleisli[F, R, *], R] {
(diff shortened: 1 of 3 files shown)


janstenpickle:
Thanks for this @bplommer, it looks like it's failing on the build unfortunately.

catostrophe:
I'll look at this on weekend.

bplommer:
Looks like I forgot about Scala 3 - I'll fix it up later.

catostrophe:
Btw, do the law checks pass  for these instances? As for me, a `Provide[F, G, R]` instance is evidence that `forall A. R => F[A] == G[A]`.

Also, I would like to look at the use-case where it's needed. I saw the PR in the `http4s` module. It is said that the client may be already lifted. In general, for me it's an antipattern to lift an F-service in a contextual effect beyond the scope where you can provide and apply the context. Doing so requires the context type to have some magic `empty` value that is provided in the outer scope, which is not always possible. That's why we have two separate effects in our implementations. What if there was no empty Span? What if Span was some opaque structure that can be instantiated only at the edge where tracing starts?

I may be absolutely wrong though. So I would like to look at the use-cases and think if it may be done in another way.

janstenpickle:
@bplommer I do agree with @catostrophe in not understanding the need for this. Would you mind explaining the use case?

bplommer:
Sorry, I'll get back to you on this soon!

catostrophe:
Closing this. It may be reopened if the motivation for this change is provided.

PR #700, opened 2022-02-14
Simplify build.sbt

bplommer:
- Removed redundant transitive module dependencies
- Removed redundant `% "test->compile"`


catostrophe:
@bplommer any research on how this may affect other repos where transitive deps might be used? Any issues should be easily fixed by providing extra deps to their builds?

bplommer:
@catostrophe I don't think this affects those at all - it's only removing dependencies that were anyway provided transitively.

catostrophe (approved):

Issue #701, opened 2022-02-14
Simplified module/package structure


bplommer:
When trialling t4c at $work I've had a lot of complaints about the number of different packages and modules that are needed for the core functionality. Here are a few tentative suggestions for reducing barriers to adoption:

### modules
- combine `model` into `kernel`
- combine  `exporter-common`, `base` and `inject` into `core`

### packages
- move contents of `model` and `inject` packages into either the base package or `trace` and `span` packages
- move contents of `kernel` package into either the base package or a `span` package
- consider changing the base package from `io.janstenpickle.trace4cats` to just `trace4cats`

Any thoughts on this?

catostrophe:
@bplommer I'm definitely for this simplification.

We're about to cut a new release soon. Actually I should have done it last week, but still have one PR open. It would be nice to include this change into the new 0.13.0 release.

The only thing that bothers me is how to make the upgrade smooth for the end users. Are you experienced with scalafix so that we could include a fix that can be applied by Scala Steward automagically?

janstenpickle:
I like the proposal too, with hindsight I went a bit far with the modularisation 😅.

I would like to get thoughts on combining `inject` and `core` though. When I first created this project I wanted a separation between the internals (span, completers and exporters) and means of injection. This was so that it could be used with Natchez without having to bring in our `inject` code and so that the standalone components (agent and collector) would not need to import the `inject` code either. What is everyone's thoughts here? Maybe keeping `inject` a separate module, but still making the proposed package changes?

> The only thing that bothers me is how to make the upgrade smooth for the end users. Are you experienced with scalafix so that we could include a fix that can be applied by Scala Steward automagically?

I agree with @catostrophe here, I think we'd need to ship some scalafix to apply the package changes, but I don't have any experience using it.

Maybe we should split this into two issues: Module consildation and package naming.

bplommer:
> I would like to get thoughts on combining `inject` and `core` though. When I first created this project I wanted a separation between the internals (span, completers and exporters) and means of injection. This was so that it could be used with Natchez without having to bring in our `inject` code and so that the standalone components (agent and collector) would not need to import the `inject` code either. What is everyone's thoughts here? Maybe keeping `inject` a separate module, but still making the proposed package changes?

How about chopping up the modules differently - moving the current `core` into `kernel` and changing `inject` to `core` – that would fit with the pattern of (for example) cats-effect, where `kernel` has the components typically needed by downstream libraries and `core` has the components needed for building an app.

[...]

PR #702, opened 2022-02-15
Consolidate modules

bplommer:
Consolidates `model` into `kernel` and `base` into `inject` - more to do

part of #701 

Depends on #700


.gitignore
```diff
@@ -1,7 +1,6 @@
 target/
 .idea
 .swp
-local.*
```
bplommer, line 4: This was causing `Local.scala` to be ignored.

janstenpickle:
@bplommer do you mind if we close this PR in favour of #742?

PR #713, opened 2022-03-26
Throwable.getMessage can return null

tdbgamer:
Default to `toString()` when `getMessage()` returns `null`.

### DIFF
--- a/modules/meta/src/main/scala/io/janstenpickle/trace4cats/meta/MetaTraceUtil.scala
+++ b/modules/meta/src/main/scala/io/janstenpickle/trace4cats/meta/MetaTraceUtil.scala
@@ -34,7 +34,7 @@ object MetaTraceUtil {
         Clock[F].realTimeInstant.flatMap { end =>
           val status = exit match {
             case ExitCase.Succeeded => SpanStatus.Ok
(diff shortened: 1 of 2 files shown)


catostrophe:
@tdbgamer does this issue cause NPE anywhere? If yes, we'll cut a patch release. If no, we'll cut a (stable) snapshot release as soon as this PR gets merged.

modules/meta/src/main/scala/io/janstenpickle/trace4cats/meta/MetaTraceUtil.scala
```diff
@@ -34,7 +34,7 @@ object MetaTraceUtil {
         Clock[F].realTimeInstant.flatMap { end =>
           val status = exit match {
             case ExitCase.Succeeded => SpanStatus.Ok
-            case ExitCase.Errored(e) => SpanStatus.Internal(e.getMessage)
+            case ExitCase.Errored(e) => SpanStatus.Internal(Option(e.getMessage).getOrElse(e.toString()))
```
catostrophe, line 37:
```suggestion
            case ExitCase.Errored(e) => SpanStatus.Internal(Option(e.getMessage).getOrElse(e.toString))
```

PR #723, opened 2022-04-06
Add helpers to construct a `Trace` backed by a `Ref`

mrdziuban:
Given a `Ref[F, Span[F]]` we can product an instance of `Local[F, Span[F]]`, which can be used to provide an implementation of `Trace[F]` via `Trace.localSpanInstance[F, F]`. This also adds supporting instances of `Lift[F, F]` and `Unlift[F, F]`.

This was inspired by [natchez's `Trace.ioTrace` method](https://github.com/tpolecat/natchez/blob/058f32553febb6b24403e765f86cbbaf3b28b08c/modules/core/shared/src/main/scala/Trace.scala#L46-L71). I was planning to add equivalents for constructing a `Trace[IO]` from an `IOLocal[Span[IO]]`, but `cats-effect` isn't currently available in the `inject` module. I think there are two approaches I could take to get it working, would you be okay with either?

1. Add `cats-effect` as a dependency to `inject`
2. Define a new `cats-effect` module that depends on `inject` and contains these helper methods

### DIFF
--- a/modules/base-laws/src/test/scala/io/janstenpickle/trace4cats/base/context/laws/LiftTests.scala
+++ b/modules/base-laws/src/test/scala/io/janstenpickle/trace4cats/base/context/laws/LiftTests.scala
@@ -1,0 +1,7 @@
+package io.janstenpickle.trace4cats.base.context.laws
+
+import io.janstenpickle.trace4cats.base.context.laws.discipline
+
+class LiftTests extends BaseSuite {
+  checkAll("Lift[Option, Option]", discipline.LiftTests[Option, Option].lift[String, Int])
+}
--- a/modules/base-laws/src/test/scala/io/janstenpickle/trace4cats/base/context/laws/UnliftTests.scala
+++ b/modules/base-laws/src/test/scala/io/janstenpickle/trace4cats/base/context/laws/UnliftTests.scala
@@ -1,0 +1,12 @@
+package io.janstenpickle.trace4cats.base.context.laws
+
+import cats.~>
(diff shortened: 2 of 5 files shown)


catostrophe:
@mrdziuban Matt, thanks for the try! Unfortunately, this implementation just does not work in a concurrent environment.

Using a global mutable variable (which is what a Ref is) being rewritten within local scopes is unsafe.  This sample would show it:
```scala
span("parent") {
  printSpanId >>
  List(
    span("child1")(printSpanParentId),
    span("child2")(printSpanParentId),
    span("child3")(printSpanParentId),
    span("child4")(printSpanParentId),
    span("child5")(printSpanParentId)
  ).parSequence_
}
```

If you run it against IO + Ref, it will spawn some child spans NOT from the parent span, which is incorrect.

If you run it against Kleisli + IO, it will work correctly.

mrdziuban:
Ahh okay, thanks for the feedback @catostrophe! Do you think an `IOLocal` solution like natchez's would work? And if so, would adding a new `cats-effect` submodule for it work?

catostrophe:
As for implementation via `IOLocal`, this would definitely work. Although you need to take into account something before you start implementing it.

We did some work to allow one to use t4c with a compound context (let's name it `Ctx`), where a `Span[F]` is just a part of it. `natchez` impl doesn't care about it. So please don't copy it as is.

I would also like to have it as a separate module since it's CE3-runtime-specific, and we depend only on ce3-kernel.

PR #729, opened 2022-04-11
Add helpers to construct `Trace` backed by `IOLocal`

mrdziuban:
Following up on #723, this adds a new `cats-effect` module with `ioLocalTrace` helper methods for constructing a `Trace[IO]` backed by an `IOLocal[Span[IO]]`

PR #736, opened 2022-04-24
Improve derivation for `TraceId.Gen` & `SpanId.Gen`

catostrophe:
I thought about @ybasket's https://github.com/trace4cats/trace4cats/pull/732#discussion_r850222888 and decided to make the derivation of `TraceId.Gen`/`SpanId.Gen` more flexible.

Now an end user can:
1. either use the default generators based on `ThreadLocalRandom`
2. or provide a different RNG instance <- _this is new_
3. or provide a fully custom generators

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -29,7 +29,6 @@ object Dependencies {
 
   lazy val caseInsensitive = "org.typelevel"           %% "case-insensitive"        % Versions.caseInsensitive
   lazy val cats = "org.typelevel"                      %% "cats-core"               % Versions.cats
-  lazy val catsEffectKernel = "org.typelevel"          %% "cats-effect-kernel"      % Versions.catsEffect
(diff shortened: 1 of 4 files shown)


project/Dependencies.scala
```diff
@@ -29,7 +29,6 @@ object Dependencies {
 
   lazy val caseInsensitive = "org.typelevel"           %% "case-insensitive"        % Versions.caseInsensitive
   lazy val cats = "org.typelevel"                      %% "cats-core"               % Versions.cats
-  lazy val catsEffectKernel = "org.typelevel"          %% "cats-effect-kernel"      % Versions.catsEffect
```
ybasket, line 32: PR #729 will need kernel, so maybe don't remove it here to avoid merge problems.

ybasket (approved):
Looks good to me!

PR #741, opened 2022-04-26
Add `Trace.WithContext`

janstenpickle:
Adds the ability to obtain a span context from a `Trace` typeclass. This has been added as a more powerful typeclass so compatibility with Natchez can be preserved.

An obivious question may be, why not just use `Ask` here? Obtaining the context this way means that we don't have to either pass another type parameter for `F` on `Span[F]` when using `Ask[G, Span[F]]` or making users manually provide an `Ask[G, SpanContext]` instance using `zoom`.

### DIFF
--- a/modules/inject/src/main/scala/io/janstenpickle/trace4cats/inject/Trace.scala
+++ b/modules/inject/src/main/scala/io/janstenpickle/trace4cats/inject/Trace.scala
@@ -12,7 +12,7 @@ import cats.syntax.option._
 import cats.syntax.show._
 import cats.{Applicative, Functor}
 import io.janstenpickle.trace4cats.base.context.{Lift, Local}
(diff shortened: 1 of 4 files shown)


modules/inject/src/main/scala/io/janstenpickle/trace4cats/inject/Trace.scala
```diff
@@ -43,6 +43,17 @@ trait Trace[F[_]] {
 }
 
 object Trace extends TraceInstancesLowPriority {
+  trait WithContext[F[_]] extends Trace[F] {
```
catostrophe, line 46: I think `IOLocal` instance also should be enhanced?

janstenpickle: Oops, forgot that PR got merged

catostrophe (approved):

PR #742, opened 2022-04-26
Module re-org

janstenpickle:
Part of #701, supersedes #702 

I've kept package names the same for now, we can probably do that in a separate PR


janstenpickle:
I think `filtering` can probably move to `core` also. WDYT @bplommer @catostrophe?

janstenpickle:
I've left `inject` intact so that people will get dependency updates. I think if we change the package name on `Trace` we can leave a `@deprecated` version in `inject`.

build.sbt
```diff
@@ -86,62 +60,60 @@ lazy val kernel =
       name := "trace4cats-kernel",
       libraryDependencies ++= Dependencies.test.map(_ % Test),
       buildInfoKeys := Seq[BuildInfoKey](version, scalaVersion, sbtVersion),
-      buildInfoPackage := "io.janstenpickle.trace4cats.kernel"
+      buildInfoPackage := "io.janstenpickle.trace4cats.kernel",
+      Test / unmanagedSourceDirectories ++= Seq(
```
catostrophe, line 64: Looks hacky. Can it be replaced with a module dependency?

build.sbt
```diff
@@ -86,62 +60,60 @@ lazy val kernel =
...
+        Dependencies.hotswapRef
+      ),
       libraryDependencies ++= Dependencies.test.map(_ % Test)
     )
-    .dependsOn(kernel, testkit % Test, `exporter-common` % Test)
+    .dependsOn(kernel, `fp-utils`, testkit % Test)
 
-lazy val base =
-  (project in file("modules/base"))
+lazy val `fp-utils` =
```
catostrophe, line 99: This triggers me the most :) The first version of tofu was named `fp-utils` which was recognized by its authors as a bad name later.

Can we find a more clear name related to contextual effects?

build.sbt
```diff
@@ -34,50 +34,24 @@ lazy val root = (project in file("."))
...
-      name := "trace4cats-model",
-      libraryDependencies ++= Seq(
-        Dependencies.catsEffectStd,
-        Dependencies.commonsCodec,
-        // Dependencies.kittens, // TODO re-add once compatible with Scala 3
-        Dependencies.caseInsensitive
-      )
-    )
-
-lazy val `model-testkit` = (project in file("modules/model-testkit"))
```
catostrophe, line 72: Will this affect other repos, where it may be used?

build.sbt
```diff
@@ -34,50 +34,24 @@ lazy val root = (project in file("."))
...
+    `fp-utils-io`,
+    `fp-utils-laws`,
     `dynamic-sampling`,
     `dynamic-sampling-config`,
-    `exporter-common`,
-    `exporter-stream`,
     filtering,
     fs2,
-    inject,
     `inject-io`,
```
bplommer, line 48: This one should probably be renamed

build.sbt
```diff
@@ -34,50 +34,24 @@ lazy val root = (project in file("."))
   .settings(noPublishSettings)
   .settings(name := "Trace4Cats")
   .aggregate(
-    base,
-    `base-io`,
-    `base-laws`,
-    core,
```
bplommer, line 40: Looks like core is missing?

build.sbt
janstenpickle: Yeah, it is hacky 😬, I did want to get your opinion on this in general:

`testkit` depends on `kernel`, but `kernel` needs some some of the arbitrary instances to do its own tests now that `core` has been moved. We could try to break up `testkit`, but there is an intractable problem if we want to avoid copy/paste.

`core` contains `RefSpanCompleter`, which is used in the `kernel` tests. I think we could move `RefSpanCompleter` into `kernel` or change testing approach by using a `WriterT` implementation.

build.sbt
janstenpickle: I don't think so, it was a bit of a hack that allowed `kernel` to be tested.

build.sbt
janstenpickle: > Do we have a written plan on package renaming? Currently, the new module organization looks inconsistent with packages and triggers my OCD.

Not yet, I wanted to get the modules moved about first. I'll create an issue for us to discuss.

janstenpickle:
> Do we have a written plan on package renaming? Currently, the new module organization looks inconsistent with packages and triggers my OCD.

No plan yet, I've created us an issue to discss #744

build.sbt
janstenpickle: I'm open to other names, this was just suggested in the original issue. Maybe just `context-utils` or `contextual-utils` is ok? I realise it still contains "utils", but unlike tofu it's a utilities library for t4c, not a standalone project.

build.sbt
janstenpickle: Yep, just fixed that

build.sbt
janstenpickle: Any suggestions, maybe something that aligns with a new package name that we'll select as part of #744?

bplommer: What do you think about just having a single `io` module that encompasses inject-io and fp-utils-io?

janstenpickle: I don't see why not, as we discussed elsewhere the only reason for having the separate `fp-utils` was because we want to keep the concerns of general purpose contextual effect manipulation separate from the rest of t4c, but that doesn't apply here

build.sbt
catostrophe: I'm fine with `context`

janstenpickle:
OK, I think that's everything in terms of module consolidation. Here's a summary of what's gone where:

`model` -> `kernel`
`core` -> `kernel`
`exporter-common` -> `core`
`exporter-stream` -> `core`
`log-exporter` -> `core`
`inject` -> `core`
`filtering` -> `core`
`rate-sampling` -> `core` & `tail-sampling`
`base` -> `context-utils`
`base-laws` -> `context-utils-laws`
`base-io` -> `io`
`inject-io` -> `io`

janstenpickle:
@bplommer @catostrophe WDYT about moving `dynamic-sampling` and `dynamic-sampling-config` into `core`?

catostrophe:
> @bplommer @catostrophe WDYT about moving `dynamic-sampling` and `dynamic-sampling-config` into `core`?

I'm fine

catostrophe (approved):

Issue #740, opened 2022-04-25
Bug: Duplicate span IDs


ybasket:
When testing #729 with my [demo repo](https://github.com/ybasket/trace4cats-iolocal-demo), I encountered many duplicate span IDs within a trace. It seems as #732 introduced a regression here when switching to `cats.effect.std.Random` – the instance acquired via `Random.javaUtilConcurrentThreadLocalRandom` seems to start with the same seed (which can't be changed) for each of CE's worker threads, resulting in similar span ID sequences being produced. This isn't well-documented, but quite visible in tests (maybe one should be added to trace4cats to check for this?). If one provides a `Random` instance backed another RNG implementation in implicit scope, the problem vanishes. Probably affects trace ID generation as well.

One can test the problem with this simple app:

```scala
import cats.effect._
import cats.effect.std.Random
import cats.syntax.all._
import io.janstenpickle.trace4cats.model.SpanId

object Randoms extends IOApp.Simple {
  override def run: IO[Unit] = (spans, spans, spans, spans).parMapN { (b1, b2, b3, b4) =>
    IO.println((b1, b2, b3, b4))
  }.flatten

  val spans = SpanId.gen[IO].replicateA(8)

  val bytesTL = IO(Random.javaUtilConcurrentThreadLocalRandom[IO].nextBytes(8).map(SpanId.unsafe).replicateA(8)).flatten
  val bytesSR = Random.scalaUtilRandom[IO].flatMap(_.nextBytes(8).map(SpanId.unsafe).replicateA(8))
  val bytesSRN = Random.scalaUtilRandomN[IO](8).flatMap(_.nextBytes(8).map(SpanId.unsafe).replicateA(8))
}
```

It'll print something like
```scala
[...]

PR #743, opened 2022-04-26
Add specs to check randomness of ID generation

ybasket:
Check that SpanIDs and TraceIds are not repeating within a small sample to detect issues with the random generator used or the resolved implicit `Gen` instances.

Proposed in https://github.com/trace4cats/trace4cats/issues/740#issuecomment-1109840582. As the upstream issue in CE still exists, two test cases fail right now. Hence marked as draft, will update once resolved.

### DIFF
--- a/modules/model/src/test/scala/io/janstenpickle/trace4cats/model/GenAssertions.scala
+++ b/modules/model/src/test/scala/io/janstenpickle/trace4cats/model/GenAssertions.scala
@@ -1,0 +1,16 @@
+package io.janstenpickle.trace4cats.model
+
+import cats.effect.IO
+import cats.kernel.Eq
+import cats.syntax.parallel._
+import org.scalatest.{Assertion, Assertions}
+
+private[model] object GenAssertions extends Assertions {
+
+  def assertAllDistinct[A: Eq](generate: IO[A]): IO[Assertion] =
+    generate
+      .parReplicateA(16) // don't choose too high or the test might sometimes fail by chance (birthday paradoxon)
+      .map { ids =>
+        assert(Eq.eqv(ids.distinct, ids), s"got only ${ids.distinct.size} distinct IDs instead of 16")
+      }
+}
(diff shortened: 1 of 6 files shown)


modules/model/src/test/scala/io/janstenpickle/trace4cats/model/GenAssertions.scala
```diff
@@ -0,0 +1,16 @@
...
+import cats.syntax.parallel._
+import org.scalatest.{Assertion, Assertions}
+
+private[model] object GenAssertions extends Assertions {
+
+  def assertAllDistinct[A: Eq](generate: IO[A]): IO[Assertion] =
+    generate
+      .parReplicateA(16) // don't choose too high or the test might sometimes fail by chance (birthday paradoxon)
+      .map { ids =>
+        assert(Eq.eqv(ids.distinct, ids), s"got only ${ids.distinct.size} distinct IDs instead of 16")
```
catostrophe, line 14: I think it's better to compare just the sizes of collections.

modules/model/src/test/scala/io/janstenpickle/trace4cats/model/SpanIdGenSpec.scala
```diff
@@ -0,0 +1,26 @@
...
+import cats.effect.unsafe.implicits.global
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+class SpanIdGenSpec extends AnyFlatSpec with Matchers {
+  behavior.of("SpanId.Gen[IO]")
+
+  it should "generate distinct SpanId instances when using ThreadLocalRandom" in {
+    val instance = implicitly[SpanId.Gen[IO]]
+    GenAssertions.assertAllDistinct(instance.gen).unsafeRunSync()
```
catostrophe, line 14: Why not just `SpanId.gen[IO]`?

catostrophe: Same below

modules/model/src/test/scala/io/janstenpickle/trace4cats/model/GenAssertions.scala
ybasket: I can change it, no problem, but may I ask why do you think it's better? We're on `List` here, so order is no issue. I only see a minor speed improvement as benefit.

modules/model/src/test/scala/io/janstenpickle/trace4cats/model/SpanIdGenSpec.scala
ybasket: No particular reason, changed it :)

catostrophe (approved):

Issue #744, opened 2022-04-27
Reworked package structure


janstenpickle:
Follows from #701. Let's discuss a new package name and structure.

- As suggested previously, we should probably go with `trace4cats` as the root package
- We'll need to provide Scalafix migrations for ScalaSteward - I've not done this before, so any help would be appreciated!
- As part of #701 and #742 some libraries will no longer be published (`inject` probably being the most used), we'll need to leave deprecated versions of interfaces to redirect to the new libraries and packages

Let's discuss!

janstenpickle:
(Will use this to track discussion)

### Kernel module

`ToHeaders` and assocateded code -> `trace4cats.kernel.headers`
`SpanCompleter, SpanExporter, SpanSampler` -> `trace4cats.kernel.export`
`*Attributes` -> `trace4cats.kernel.attributes`

(Maybe we should just have `trace4cats` instead of `trace4cats.kernel`?)

### Core module

`LogSpanCompleter, LogSpanExporter` -> `trace4cats.export.log`
`inject.*` -> ???
`rate.*` -> ???

bplommer:
> (Maybe we should just have trace4cats instead of trace4cats.kernel?)

👍 cats and cats-effect have kernel stuff in a kernel package but aliased to the root package - not sure why but might be worth looking into the reasoning

bplommer:
`trace4cats.base.context` -> `trace4cats.mtl`?

janstenpickle:
> `trace4cats.base.context` -> `trace4cats.mtl`?

Maybe just `trace4cats.context` as it's not a full MTL, it just deals with contextual effects

janstenpickle:
> 👍 cats and cats-effect have kernel stuff in a kernel package but aliased to the root package - not sure why but might be worth looking into the reasoning

So I've found out why this is; they want to keep the kernel code in its own package and library, but also enable app developers to import `cats.effect._` and have everything they need. I quite like that approach as it will allow us to organise code a nice package structure, while supporting a wildcard import.

catostrophe:
Wow
https://github.com/scala-steward-org/scala-steward/issues/2327

catostrophe:
We'll have to add rewrite rules to
https://github.com/scala-steward-org/scala-steward/blob/main/modules/core/src/main/resources/scalafix-migrations.conf

janstenpickle:
[...]

PR #748, opened 2022-04-30
Migrate classes to new package structure

janstenpickle:
Addresses #744


modules/tail-sampling/src/main/scala/trace4cats/samplling/tail/RateTailSpanSampler.scala
```diff
@@ -1,12 +1,11 @@
-package io.janstenpickle.trace4cats.rate.sampling
+package trace4cats.samplling.tail
```
catostrophe, line 1: Typo

modules/meta/src/test/scala/io/janstenpickle/trace4cats/meta/TracedSpanExporterSpec.scala
```diff
@@ -7,13 +7,13 @@ import cats.effect.unsafe.implicits.global
 import fs2.{Chunk, Stream}
```
catostrophe, line 7: this file should be changed

janstenpickle: Yep, this whole branch is still WIP

catostrophe:
@janstenpickle do you need help with this PR?

janstenpickle:
@bplommer how does this new structure look to you?

catostrophe:
Will review tonight

modules/context-utils-laws/src/main/scala/trace4cats/context/laws/discipline/AskTests.scala
```diff
@@ -14,15 +14,15 @@
  * limitations under the License.
  */
 
-package io.janstenpickle.trace4cats.base.context
-package laws
-package discipline
+package trace4cats.context.laws.discipline
```
bplommer, line 17: Unrelated but might be a good time to address it - it's better for library tests to use a *different* root package from the code under test, to better reflect visibility and implicit resolution for real user code.
```suggestion
package trace4catsTests.context.laws.discipline
```

modules/core/src/main/scala/trace4cats/package.scala
```diff
@@ -0,0 +1,30 @@
+import cats.data.Kleisli
+import cats.effect.kernel.Resource
+import trace4cats.model.{SpanKind, TraceHeaders}
+import trace4cats.{kernel => t4ckernel}
+
+import scala.annotation.unchecked.uncheckedVariance
+
+package object trace4cats {
+  type SpanName = String
```
bplommer, line 9: Unrelated: let's make this a value class instead of a type alias. But doesn't have to be now.

modules/core/src/main/scala/trace4cats/package.scala
```diff
@@ -0,0 +1,30 @@
...
+  type SpanParams = (SpanName, SpanKind, TraceHeaders, ErrorHandler)
+  type ResourceKleisli[F[_], -In, +Out] = Kleisli[Resource[F, +*], In, Out @uncheckedVariance]
```
bplommer, line 11: I was going to suggest renaming `ResourceKleisli` to something that reflects purpose rather than implementation, and possibly changing it from a type alias to part of a class heirarchy with `EntryPoint`. But again, not something that should happen in this PR - just mentioning it while I remember

bplommer:
Looks good! I wonder if types from `model` should also be aliased in the main package object?

modules/context-utils-laws/src/main/scala/trace4cats/context/laws/discipline/AskTests.scala
janstenpickle: Oh good to know, thanks

modules/core/src/main/scala/trace4cats/package.scala
janstenpickle: This is something I considered when implementing this, I didn't want to burden the user with having to wrap every span name. We could have implicit conversions for strings to the value class, but then there is no obligation on the user to pass the _right_ string. It would be useful for internal library devs but an implicit conversion might create a bit of a footgun here.

Not sure how strongly you feel about this or if theres a really compelling reason for using a value class, I just wanted to lay out why this hasn't been done before.

modules/core/src/main/scala/trace4cats/package.scala
janstenpickle: The main reason for not making this its own class is that we get a lot of things for free with `Kleisli` and its input and output aren't fixed to the input/output of the entry point. This allows us to compose the entrypoint with other means of establishing some context from some input. There's an example of this in the[ http4s library](https://github.com/trace4cats/trace4cats-http4s/blob/master/modules/http4s-server/src/main/scala/io/janstenpickle/trace4cats/http4s/server/Http4sResourceKleislis.scala).

modules/core/src/main/scala/trace4cats/package.scala
bplommer: I don't feel super strongly about it, I mainly just dislike the illusory type safety provided by type aliases - what do you think about just using `String` explicitly?

janstenpickle:
> Looks good! I wonder if types from `model` should also be aliased in the main package object?

Yes, that might be handy for things like the IDs and context, I'll take a look.

modules/core/src/main/scala/trace4cats/package.scala
janstenpickle: That's fair, it was added so that it could be easily identified in a tuple passed to `ResourceKleisli`, however this could be done using a case class as input rather than a tuple. I'll create an issue to do that and remove the type alias.

janstenpickle: #757

janstenpickle:
@bplommer @catostrophe how does this look to you both now, shall I merge?

I've got a scalafix branch in progress at the moment as well, so once the other repos are done we should be getting close to a release

catostrophe:
Sorry for being so slow, been busy at $WORK. Checking now...

janstenpickle:
> Sorry for being so slow, been busy at $WORK. Checking now...

No problem, I know the feeling!

modules/context-utils-laws/src/main/scala/trace4catstests/context/laws/discipline/UnliftTests.scala
```diff
@@ -1,11 +1,11 @@
-package io.janstenpickle.trace4cats.base.context
-package laws
-package discipline
+package trace4catstests.context.laws.discipline
```
catostrophe, line 1: Is it intended? 🤔

modules/context-utils-laws/src/test/scala/trace4cats/context/laws/IdTests.scala
```diff
@@ -1,10 +1,10 @@
...
+package trace4cats.context.laws
 
-import cats.~>
 import cats.laws.discipline.SerializableTests
-import io.janstenpickle.trace4cats.base.context.Unlift
-import io.janstenpickle.trace4cats.base.context.laws.discipline.UnliftTests
+import cats.~>
 import org.scalacheck.Cogen
+import trace4cats.context.Unlift
+import trace4catstests.context.laws.discipline.UnliftTests
```
catostrophe, line 7: Same here

modules/iolocal/src/test/scala/trace4cats/context/iolocal/laws/IOLocalTests.scala
```diff
@@ -1,15 +1,15 @@
...
-import io.janstenpickle.trace4cats.base.context.iolocal.GenIO
-import io.janstenpickle.trace4cats.base.context.iolocal.instances._
-import io.janstenpickle.trace4cats.base.context.laws.BaseSuite
-import io.janstenpickle.trace4cats.base.context.laws.discipline.ProvideTests
 import org.scalacheck.{Arbitrary, Cogen, Gen}
+import trace4cats.context.Provide
+import trace4cats.context.iolocal.GenIO
+import trace4cats.context.iolocal.instances.ioLocalProvide
+import trace4cats.context.laws.BaseSuite
+import trace4catstests.context.laws.discipline.ProvideTests
```
catostrophe, line 12: Same here

catostrophe (commented):
Why `trace4catstests`?

catostrophe:
> > Why `trace4catstests`?
> 
> @catostrophe - It was suggested by @bplommer. Sorry, I resolved [the comment](https://github.com/trace4cats/trace4cats/pull/748#discussion_r879241965) so you wouldn't have seen it!

I still haven't got the idea behind it.

I've never seen such a separation before in the context of `discipline` and law tests. Instead, laws and tests are usually in the same namespace and are close enough, because, if they are ever edited, they are edited together.

See: 
https://github.com/typelevel/cats/tree/main/kernel-laws/shared/src/main/scala/cats/kernel/laws/discipline
https://github.com/typelevel/cats-tagless/tree/master/laws/src/main/scala/cats/tagless/laws/discipline
https://github.com/typelevel/cats-effect/tree/series/3.x/laws/shared/src/main/scala/cats/effect/laws

catostrophe:
These "tests" are actually rule setst that you use in real tests to check some concrete implementation for satisfying them.

bplommer:
> I've got to admit I took the recommendation at face value. I wonder if you were thinking more of the separate test modules rather than discipline tests @bplommer?

Yes, afraid I didn't look very closely before suggesting that. Anything that's going to be published as an artifact, I'd leave in the `trace4cats` package.

catostrophe:
Could we revert the introduction of `trace4catstests` package? 

The rest is fine 👍

janstenpickle:
> Could we revert the introduction of `trace4catstests` package?
> 
> The rest is fine 👍

Cool, I'll revert this now then merge. Thanks all!

PR #752, opened 2022-05-07
Rename io packages to iolocal

ybasket:
Rename packages as "io" is a common first segment of package names (`io.circe` for example) and can hence be very annoying to users that then have to sort their imports or prefix them with `_root_`.

Open question: Shall the module also be renamed from `io` to `iolocal` to keep naming consistent?


catostrophe:
Yes, let's rename the package. I thought about it.

catostrophe (approved):

PR #774, opened 2022-07-16
Add scalafix rules for package migrations

janstenpickle:
Bit of a brute force approach, but it seems to work 🤷


catostrophe:
Wow!

janstenpickle:
I've opted to move these into a separate repo seeing as they cover multiple dependencies: https://github.com/trace4cats/trace4cats-scalafix/pull/1

PR #816, opened 2022-11-10
Rework span queued span completer

janstenpickle:
The queued span completer was causing contention around the `Ref` for
recording in flight requests. Seeing as this was only really used to
ensure that the completer can be shut down properly, there's a better
way of doing this new we have CE3.

### DIFF
--- a/modules/core-tests/src/test/scala/trace4cats/QueuedSpanCompleterSpec.scala
+++ b/modules/core-tests/src/test/scala/trace4cats/QueuedSpanCompleterSpec.scala
@@ -19,7 +19,7 @@ class QueuedSpanCompleterSpec extends AnyFlatSpec with Matchers with TestInstanc
   behavior.of("QueuedSpanCompleter")
 
   def delayedExporter(ref: Ref[IO, Int]): SpanExporter[IO, Chunk] = new SpanExporter[IO, Chunk] {
-    def exportBatch(batch: Batch[Chunk]): IO[Unit] = IO.sleep(10.millis) >> ref.update(_ + batch.spans.size)
+    def exportBatch(batch: Batch[Chunk]): IO[Unit] = IO.sleep(20.millis) >> ref.update(_ + batch.spans.size)
   }
 
   def refExporter(ref: Ref[IO, Chunk[CompletedSpan]]): SpanExporter[IO, Chunk] = new SpanExporter[IO, Chunk] {
@@ -44,7 +44,7 @@ class QueuedSpanCompleterSpec extends AnyFlatSpec with Matchers with TestInstanc
     } yield res
 
     val result = test.unsafeRunSync()
-    result shouldBe (<(10.millis))
+    result shouldBe (<(20.millis))
   }
 
   it should "export all spans on shutdown" in forAll { (builder: CompletedSpan.Builder) =>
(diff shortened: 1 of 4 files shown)


modules/core-tests/src/test/scala/trace4cats/QueuedSpanCompleterSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+  it should "export all spans on shutdown" in forAll { (builder: CompletedSpan.Builder) =>
+    val test = for {
+      ref <- Ref.of[IO, Chunk[CompletedSpan]](Chunk.empty)
+      exporter = refExporter(ref)
+      _ <- QueuedSpanCompleter[IO](
+        TraceProcess("completer-test"),
+        exporter,
+        CompleterConfig(bufferSize = 50000, batchSize = 200)
+      ).use { completer =>
+        List.fill(10000)(completer.complete(builder)).sequence
```
TimWSpence, line 66: Is it worth doing something here to ensure that these are definitely not all handled in one `groupWithin`? Or a separate test for that? Of course it _should_ all be `uncancelable` but it would be nice to be sure that the tests covered that

modules/core-tests/src/test/scala/trace4cats/QueuedSpanCompleterSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+      random: Random[IO] <- Random.scalaUtilRandom[IO]
+      exporter = delayedExporter(ref)
+      res <- QueuedSpanCompleter[IO](
+        TraceProcess("completer-test"),
+        exporter,
+        CompleterConfig(bufferSize = 5, batchSize = 1)
+      ).use { completer =>
+        val randomRequestTime = for {
+          rand <- random.betweenDouble(0, 1)
+          _ <- IO.sleep((1.second * rand).asInstanceOf[FiniteDuration])
```
TimWSpence, line 46: Is this just jitter so we don't hammer it with everything at once?

TimWSpence (approved):

modules/core/src/main/scala/trace4cats/QueuedSpanCompleter.scala
```diff
@@ -22,50 +20,52 @@ object QueuedSpanCompleter {
...
-        )
-
-        inFlight.get
-          .map(_ == realBufferSize)
-          .ifM(warnLog, enqueue >> hasLoggedWarn.set(false))
-      }
+    for {
+      channel <- Resource.eval(Channel.bounded[F, CompletedSpan](realBufferSize))
+      errorQueue <- Resource.eval(Queue.bounded[F, Either[Channel.Closed, Boolean]](1))
+      _ <- Stream
```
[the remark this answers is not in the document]
janstenpickle: Not really, I've just added the log for completeness

modules/core-tests/src/test/scala/trace4cats/QueuedSpanCompleterSpec.scala
janstenpickle: Yep, I believe so, I think @sgjbryan wanted to improve this a bit

modules/core-tests/src/test/scala/trace4cats/QueuedSpanCompleterSpec.scala
sgjbryan: Just to be on the safe side, I'd suggest we might want to the number here to not be divided cleanly into batches (so `10001` for example). That way we verify that incomplete/partial batches are handled cleanly during shutdown

modules/core-tests/src/test/scala/trace4cats/QueuedSpanCompleterSpec.scala
janstenpickle: sounds good, I just removed it

modules/core-tests/src/test/scala/trace4cats/QueuedSpanCompleterSpec.scala
janstenpickle: Great idea, I'll change the tests

PR #850, opened 2023-03-01
Make parsing of Google's X-Cloud-Trace-Context less strict

sideeffffect:
The `;o=...` part is supposed to be optional

### DIFF
--- a/modules/kernel/src/main/scala/trace4cats/kernel/headers/GoogleCloudTraceToHeaders.scala
+++ b/modules/kernel/src/main/scala/trace4cats/kernel/headers/GoogleCloudTraceToHeaders.scala
@@ -50,7 +50,7 @@ private[trace4cats] object GoogleCloudTraceToHeaders {
         traceId = traceId,
         spanId = spanId,
         parent = none,
-        traceFlags = TraceFlags(if (enabled == Some("1")) SampleDecision.Include else SampleDecision.Drop),
+        traceFlags = TraceFlags(if (enabled == Some("0")) SampleDecision.Drop else SampleDecision.Include),
         traceState = TraceState.empty,
         isRemote = true
       )
--- a/modules/kernel/src/main/scala/trace4cats/kernel/headers/GoogleCloudTraceToHeaders.scala
+++ b/modules/kernel/src/main/scala/trace4cats/kernel/headers/GoogleCloudTraceToHeaders.scala
@@ -50,8 +50,8 @@ private[trace4cats] object GoogleCloudTraceToHeaders {
         spanId = spanId,
         parent = none,
         traceFlags = TraceFlags(Option(enabled) match { // `enabled` can be `null`
-          case Some("0") => SampleDecision.Drop
-          case _ => SampleDecision.Include
+          case Some("1") | None => SampleDecision.Include
+          case _ => SampleDecision.Drop
         }),
         traceState = TraceState.empty,
         isRemote = true
--- a/modules/kernel/src/main/scala/trace4cats/kernel/headers/GoogleCloudTraceToHeaders.scala
+++ b/modules/kernel/src/main/scala/trace4cats/kernel/headers/GoogleCloudTraceToHeaders.scala
@@ -8,6 +8,8 @@ import org.typelevel.ci._
 import trace4cats.kernel.ToHeaders
 import trace4cats.model._
 
+import scala.Option
+
 private[trace4cats] class GoogleCloudTraceToHeaders extends ToHeaders {
   import GoogleCloudTraceToHeaders._
 
@@ -38,8 +40,7 @@ private[trace4cats] object GoogleCloudTraceToHeaders {
       |""".stripMargin.r
 
   def parse(header: String): Either[Throwable, SpanContext] = header match {
-    case headerPattern(traceId, spanId, enabled0) =>
-      val enabled = Option(enabled0)
+    case headerPattern(traceId, spanId, enabled) =>
       for {
         traceId <- Either.fromOption(TraceId.fromHexString(traceId), new Exception("invalid trace ID"))
         spanId <- Either.fromOption(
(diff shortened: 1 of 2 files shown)


sideeffffect:
CCing also @timwspence :pray:

modules/kernel/src/main/scala/trace4cats/kernel/headers/GoogleCloudTraceToHeaders.scala
```diff
@@ -47,7 +50,7 @@ private[trace4cats] object GoogleCloudTraceToHeaders {
         traceId = traceId,
         spanId = spanId,
         parent = none,
-        traceFlags = TraceFlags(if (enabled == "1") SampleDecision.Include else SampleDecision.Drop),
+        traceFlags = TraceFlags(if (enabled == Some("1")) SampleDecision.Include else SampleDecision.Drop),
```
bambuchaAdm, line 53: Here "0" means drop - absence means use `SpanSampler` to make decision.

dpevunov-cp: How is it working in the original code then? I dont see it being set to None

dpevunov-cp: @sideeffffect do we really want to change the old behaviour where it is always SampleDecision.Drop except for explicitly set "1"?

sideeffffect: The old behavior allowed only for `1` and `0`.

Me and @bambuchaAdm agreed that it's sensible to drop only if explicitly requested (using `0`). Otherwise keep the trace.

modules/kernel/src/main/scala/trace4cats/kernel/headers/GoogleCloudTraceToHeaders.scala
```diff
@@ -28,26 +28,31 @@ private[trace4cats] object GoogleCloudTraceToHeaders {
...
-          SpanId.fromHexString("%016x".format(BigInt(spanId))),
+          Either.catchNonFatal(BigInt(spanId)).toOption.flatMap(bi => SpanId.fromHexString("%016x".format(bi))),
           new Exception("invalid span ID")
         )
       } yield SpanContext(
         traceId = traceId,
         spanId = spanId,
         parent = none,
-        traceFlags = TraceFlags(if (enabled == "1") SampleDecision.Include else SampleDecision.Drop),
+        traceFlags = TraceFlags(Option(enabled) match { // `enabled` can be `null`
```
ybasket, line 52: The change here swaps the default for unexpected values from dropping to sampling – I think it would be better to do `case Some("1") | None => SampleDecision.Include` instead.

modules/kernel/src/main/scala/trace4cats/kernel/headers/GoogleCloudTraceToHeaders.scala
```diff
@@ -28,26 +28,31 @@ private[trace4cats] object GoogleCloudTraceToHeaders {
   // from https://cloud.google.com/trace/docs/setup
   val headerPattern =
     """(?xi)
-      |([0-9a-f]+) # trace ID
+      |([^\/]+)    # trace ID
```
ybasket, line 31: What motivates this change and the one below? Can the trace ID ever been non-hexadecimal or the span ID non-decimal?

modules/kernel/src/main/scala/trace4cats/kernel/headers/GoogleCloudTraceToHeaders.scala
ybasket: Oh, just saw there was already a discussion on that – I'm rather in the explicit sampling camp ⛺

modules/kernel/src/main/scala/trace4cats/kernel/headers/GoogleCloudTraceToHeaders.scala
sideeffffect: You will get error in either case. But this way you get more precise error. The regex should try to match only the rough format of the string.

modules/kernel/src/main/scala/trace4cats/kernel/headers/GoogleCloudTraceToHeaders.scala
sideeffffect: This sounds reasonable, I switched to `case Some("1") | None => SampleDecision.Include`

modules/kernel/src/main/scala/trace4cats/kernel/headers/GoogleCloudTraceToHeaders.scala
```diff
@@ -28,26 +28,31 @@ private[trace4cats] object GoogleCloudTraceToHeaders {
...
       |""".stripMargin.r
 
   def parse(header: String): Either[Throwable, SpanContext] = header match {
     case headerPattern(traceId, spanId, enabled) =>
       for {
         traceId <- Either.fromOption(TraceId.fromHexString(traceId), new Exception("invalid trace ID"))
         spanId <- Either.fromOption(
-          SpanId.fromHexString("%016x".format(BigInt(spanId))),
+          Either.catchNonFatal(BigInt(spanId)).toOption.flatMap(bi => SpanId.fromHexString("%016x".format(bi))),
           new Exception("invalid span ID")
```
janstenpickle, line 46: The parse method supports surfacing an exception on the left of an either so I'd suggest doing this instead:

`Either.catchNonFatal(BigInt(spanId)).flatMap(bi => Either.fromOption(SpanId.fromHexString("%016x".format(bi)), new Exception("invalid span ID")))`

janstenpickle (commented):
Thanks for this @sideeffffect, just had one nitpick on parsing the BigInt error handling, other than that happy to merge!

modules/kernel/src/main/scala/trace4cats/kernel/headers/GoogleCloudTraceToHeaders.scala
sideeffffect: Done :+1:

PR #888, opened 2023-07-18
Fix shutdown of QueuedSpanCompleter

ybasket:
By having an uncancelable fiber via background and discarding the reference to it, the shut down wasn't awaiting the sending process to finish, causing spans to be lost. This bug probably got amplified in newer CE versions with improved performance, it manifested regularly in the Jaeger-based tests, see https://github.com/trace4cats/trace4cats-jaeger/pull/175.

I tried adding a test case that would reproduce the issue, but failed as it seems to require real concurrency with thread switches that are hard to set up in a simple test case where as they come naturally from network I/O in integration tests.

### DIFF
--- a/modules/core/src/main/scala/trace4cats/QueuedSpanCompleter.scala
+++ b/modules/core/src/main/scala/trace4cats/QueuedSpanCompleter.scala
@@ -57,14 +57,15 @@ object QueuedSpanCompleter {
         .compile
         .drain
         .background
-      _ <- exportBatches(channel.stream).uncancelable.background
+      exportFiber <- exportBatches(channel.stream).uncancelable.background
         .onFinalize(Logger[F].info("Shut down queued span completer"))
+      _ <- Resource.onFinalize(exportFiber.void) // join fiber to ensure everything was sent
       _ <- Resource.onFinalize(channel.close.void)
     } yield new SpanCompleter[F] {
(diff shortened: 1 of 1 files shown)


modules/core/src/main/scala/trace4cats/QueuedSpanCompleter.scala
```diff
@@ -57,14 +57,15 @@ object QueuedSpanCompleter {
...
-      _ <- exportBatches(channel.stream).uncancelable.background
+      exportFiber <- exportBatches(channel.stream).uncancelable.background
         .onFinalize(Logger[F].info("Shut down queued span completer"))
+      _ <- Resource.onFinalize(exportFiber.void) // join fiber to ensure everything was sent
       _ <- Resource.onFinalize(channel.close.void)
     } yield new SpanCompleter[F] {
       override def complete(span: CompletedSpan.Builder): F[Unit] =
         channel
           .trySend(span.build(process))
-          .flatMap(errorQueue.tryOffer(_).start.void)
```
ybasket, line 67: This is not strictly needed to fix the issue, but as `tryOffer` is guaranteed to never semantically block, I don't see why we couldn't save ourselves from the overhead of starting a fiber on this relatively hot path.

gbastkowski (approved):

PR #893, opened 2023-08-11
Fix module dependencies for iolocal

ybasket:
By using an unfortunate sbt dependency projection, the iolocal module pulled test libs like scalacheck into the compile scope and made apps depending on iolocal have those on their runtime classpath.


janstenpickle (approved):
Ooop, nice catch @ybasket

PR #915, opened 2023-10-29
Fix span completer freezing issue

kovstas:
`onError` in the `QueuedSpanCompleter.exportBatches` was replaced with `handleErrorWith` not to fail the source stream otherwise the `SpanCompleter` will receive spans but will not be able to export them.


ybasket (approved):

ybasket:
Thank you! Released in https://github.com/trace4cats/trace4cats/releases/tag/v0.14.7

